PR #1, opened 2021-03-21
Dockerfile for Mirroring Agent

A0su:
Appears to exhibit the same behavior as local process. Any feedback is appreciated!

### DIFF
--- a/mirroring/Dockerfile
+++ b/mirroring/Dockerfile
@@ -1,13 +1,9 @@
 FROM golang:latest
 
 RUN mkdir /mirroring
-RUN apt-get install curl
-RUN apt-get install openssl
 WORKDIR /mirroring
 
-COPY ./mirror_log.go .
-COPY ./go.mod .
-COPY ./go.sum .
+COPY ./mirror_log.go ./go.mod ./go.sum ./
 
 RUN go mod download
 
--- a/mirroring/Dockerfile
+++ b/mirroring/Dockerfile
@@ -7,10 +7,9 @@ COPY ./mirror_log.go ./go.mod ./go.sum ./
 
 RUN go mod download
 
-RUN GO111MODULE=on GOOS=linux go build -o mirroring .
+RUN GO111MODULE=on GOOS=linux GOARCH=amd64 CGO_ENABLED=0 go build -o mirroring .
 
-from ubuntu:latest
-WORKDIR /root/
-COPY --from=0 /mirroring/mirroring .
-RUN apt-get -y update && apt-get -y upgrade ca-certificates
-CMD ["./mirroring"]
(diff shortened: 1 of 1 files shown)


mirroring/Dockerfile
```diff
@@ -0,0 +1,20 @@
+FROM golang:latest
+
+RUN mkdir /mirroring
+RUN apt-get install curl
+RUN apt-get install openssl
+WORKDIR /mirroring
+
+COPY ./mirror_log.go .
```
SantiagoTorres, line 8: Can we do this in a single copy call ? so as to do this in a single layer.

mirroring/Dockerfile
```diff
@@ -0,0 +1,20 @@
+FROM golang:latest
+
+RUN mkdir /mirroring
+RUN apt-get install curl
+RUN apt-get install openssl
```
SantiagoTorres, line 5: I don't think we would have to install these anymore

SantiagoTorres (changes requested):
Some small nits

A0su:
Pushed changes to clean up the Dockerfile as the comments suggested.

mirroring/Dockerfile
```diff
@@ -0,0 +1,16 @@
...
+
+RUN go mod download
+
+RUN GO111MODULE=on GOOS=linux go build -o mirroring .
+
+from ubuntu:latest
+WORKDIR /root/
+COPY --from=0 /mirroring/mirroring .
+RUN apt-get -y update && apt-get -y upgrade ca-certificates
+CMD ["./mirroring"]
```
SantiagoTorres, line 16: Curious, if you use alpine/nix instead of ubuntu, would it work without having this RUN command? (it just happens to make the container non-reproducible)

A0su: I was able to get the Dockerfile to work with Alpine w/o the RUN command I just ended up having to change the compile flags. It looks like this:
 
```
FROM golang:latest

RUN mkdir /mirroring
WORKDIR /mirroring

COPY ./mirror_log.go ./go.mod ./go.sum ./

RUN go mod download

RUN GO111MODULE=on GOOS=linux GOARCH=amd64 CGO_ENABLED=0 go build -o mirroring .

from alpine:latest
WORKDIR ./
COPY --from=0 /mirroring/mirroring ./mirroring
ENTRYPOINT ["./mirroring"]
```

Should we make the switch to this version of the Dockerfile?

SantiagoTorres: LGTM! I susepct the image is also smaller.

A0su: Yup! Nearly 5 times smaller

PR #19, opened 2021-04-19
WIP: Log handler init

efebarlas:
This request introduces the log handler object, allowing for an SDK-like interface for developers that want to make monitors themselves.


mirroring/handler.go
```diff
@@ -0,0 +1,230 @@
...
+	"os"
+
+	"github.com/go-openapi/strfmt"
+	trilliantypes "github.com/google/trillian/types"
+	"github.com/sigstore/rekor/pkg/generated/client"
+	"github.com/sigstore/rekor/pkg/generated/models"
+	"github.com/spf13/viper"
+)
+
+// consider adding a rekorClient field so that the same client struct is used for all operations, potentially saving some time
```
SantiagoTorres, line 19: is this a FIXME/TODO? can we label it as such? we could also turn it into a ticket.

mirroring/handler.go
```diff
@@ -0,0 +1,230 @@
...
+type LogHandler struct {
+	metadata        TreeMetadata
+	client          *client.Rekor
+	newLeavesBuffer []Artifact
+}
+
+func LoadFromRemote(serverURL string) (h LogHandler, err error) {
```
SantiagoTorres, line 26: Not entirely sure what this API funcion gives me. Is this part of a CLI entrypoint to load a remote MHT? is that happening automagically on `GetLogInfo` ? or on `h.client` ?

mirroring/handler.go
```diff
@@ -0,0 +1,230 @@
...
+	metadata, err := LoadTreeMetadata()
+	if err != nil {
+		return handler, err
+	}
+	handler.client, err = NewClient()
+	if err != nil {
+		return handler, err
+	}
+	handler.metadata = metadata
+	// stored information will contain duplicates. should these duplicates be checked?
```
SantiagoTorres, line 70: Can we ticketize this?

mirroring/handler.go
```diff
@@ -0,0 +1,230 @@
...
+	*h.metadata.LogInfo.SignedTreeHead.Signature = sig
+}
+
+func (h *LogHandler) SetLocalTreeSize(treeSize int64) {
+	h.metadata.SavedMaxIndex = treeSize - 1
+}
+
+// if kind=="", get all kinds of leaves
+func (h *LogHandler) GetAllLeavesForKind(kind string) error {
+	str := viper.GetString("tree_file_dir")
```
SantiagoTorres, line 177: I think we could call this variable something more descriptive. What about `tree_file_dir_path` ? :P

mirroring/handler.go
```diff
@@ -0,0 +1,230 @@
...
+				return err
+			}
+		}
+
+		err = json.Unmarshal([]byte(line), &leaf)
+		if err != nil {
+			return err
+		}
+		artifacts = append(artifacts, leaf)
+	}
```
SantiagoTorres, line 223: It's not immediately obvious why we need to split these into two for loops instead of using the kind statement from within (as a filter):

`if leaf.Kind == kind || kind == ""`

or somesuch, what do you think?

SantiagoTorres (commented):
Some small comments to get started. I think I'll do a deeper dive tomorrow/wed.

mirroring/handler.go
efebarlas: This was a TODO, but it's already done, so I deleted the message.

mirroring/handler.go
efebarlas: I thought that the two main construction scenarios for the log-handler would be to either:
-Fill in the public key, root hash, tree size and other metadata from a previously stored state
or
-Given a provided URL, to fetch the metadata from a Rekor log.

For the second case (LoadFromRemote), GetLogInfo is called to fetch the metadata and store it in the log-handler, and then a Rekor client is initiated so that for almost all other operations where a Rekor API call is needed, the client is not instantiated again and again, allowing for a gain in performance.

mirroring/handler.go
efebarlas: I removed the comment and added the checks in the next commit.

mirroring/handler.go
efebarlas: That makes sense, I renamed every '[prefix]_dir' key to '[prefix]_directory'

mirroring/handler.go
efebarlas: I split them for the reason that if kind == "", then separating them would allow the for loop not have to go through the condition check every time before an artifact is appended, making the code run a bit faster.

On second thought however, compiler optimization and branch prediction probably makes this split pointless, so I combined them in the next commit.

efebarlas:
I have noticed some test cases were commented out because they were failing. I will be fixing those and then after a review, we can merge this.

haydentherapper:
Closing older PRs, feel free to reopen if you want to continue this work.

PR #24, opened 2021-05-03
Tekton-Deployment

gonza495:
This is what Luke and I had been working on. This deploys Tekton and makes sure sbin is in PATH


efebarlas:
Can you also add your signature so that the DCO action succeeds?

haydentherapper:
Going to focus on making this usable via GitHub Actions.

PR #27, opened 2021-09-07
Add root hash consistency check

hojoung97:
I added a very basic root hash consistency check.
1. Grabs the latest log consistency proof from Rekor server
2. Checks if the last seen root hash is part of the latest log consistency proof from step 1

For now, I have simply queried the latest log information to check it against the latest log consistency proof.
Moving forward we can probably temporarily use a text file with latest root hash saved until we actually run the monitor's transparency log that will hold the latest root hash audited.

### DIFF
--- a/mirroring/mirroring.go
+++ b/mirroring/mirroring.go
@@ -40,6 +40,8 @@ import (
 	"github.com/spf13/viper"
 )
 
+const rekorServerURL = "https://api.sigstore.dev"
+
 type getCmdOutput struct {
 	Body           types.EntryImpl
 	LogIndex       int
@@ -59,15 +61,7 @@ type queueElement struct {
 	depth int64
 }
 
-func GetPublicKey() (string, error) {
-	// Initialize Rekor client
-	rekorServerURL := viper.GetString("rekorServerURL")
-	rekorClient, err := client.GetRekorClient(rekorServerURL)
-	if err != nil {
-		return "", err
-	}
-
-	// Get Rekor public key
+func GetPublicKey(rekorClient *gclient.Rekor) (string, error) {
 	pubkeyResp, err := rekorClient.Pubkey.GetPublicKey(nil)
 	if err != nil {
 		return "", err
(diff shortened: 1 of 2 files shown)


mirroring/mirroring.go
```diff
@@ -90,6 +91,42 @@ func GetLogInfo() (*models.LogInfo, error) {
 	return logInfoResp.GetPayload(), nil
 }
 
+func GetLogProof(firstSize *int64) (*models.ConsistencyProof, error) {
+	rekorServerURL := viper.GetString("rekorServerURL")
+	rekorClient, err := client.GetRekorClient(rekorServerURL)
```
SantiagoTorres, line 96: Hmm, I wonder if we'd like to split out the CLI/config setting parse routines from this function and make the server url a parameter instead...

mirroring/mirroring.go
```diff
@@ -90,6 +91,42 @@ func GetLogInfo() (*models.LogInfo, error) {
...
+	params.LastSize = *logInfo.TreeSize
+
+	logProofResp, err := rekorClient.Tlog.GetLogProof(params)
+	if err != nil {
+		return nil, err
+	}
+	return logProofResp.GetPayload(), nil
+}
+
+func VerifyConsistencyProof(rootHash *string, logProof *models.ConsistencyProof) bool {
```
SantiagoTorres, line 119: I'm not sure if this function is doing all it should, but we can discuss this. Shouldn't it be reconstructing the new roothash using the old hash and the proof

SantiagoTorres (changes requested):
I think there's some bits that I'd like to discuss, but this is looking in the right direction!

mirroring/mirroring.go
hojoung97: Yes, I had somewhat related doubt which is the fact that we are initializing a new rekor client for each operation we perform. We can make the server URL as a parameter but what do you think of making a rekor client instance as a parameter?

SantiagoTorres: Hmm, good point. Let's do the latter :+1:

hojoung97:
I made some changes here.
I want to just highlight that the root hash consistency check is now computing a new root hash from last seen root hash and compares computed hash against the newly advertised root hash. 
For now, I inserted a very basic testing with the last tree size being 1.

SantiagoTorres (approved):
Fantastic work!

SantiagoTorres:
FWIW I liked your use of the rfc6962 module. IT keeps things clear and succint :)

PR #28, opened 2021-09-16
Add inclusion check and fix consistency check

hojoung97:
Signed-off-by: Hojoung Jang <jang88@purdue.edu>

#### Summary
- Add a function that performs inclusion check of a given entry
- Fix wrong implementation of root hash computation
  - Previously, the order of concatenating two hashes was not considered 
    ex. `hash(left child hash + right child hash) != hash(right child hash + left child hash)`
  - Used Trillian's implementation which does consider this

### DIFF
--- a/mirroring/mirroring.go
+++ b/mirroring/mirroring.go
@@ -26,6 +26,7 @@ import (
 	"errors"
 
 	"github.com/go-openapi/runtime"
+	"github.com/google/trillian/merkle/logverifier"
 	rfc6962 "github.com/google/trillian/merkle/rfc6962/hasher"
 	"github.com/sigstore/rekor/pkg/client"
 	gclient "github.com/sigstore/rekor/pkg/generated/client"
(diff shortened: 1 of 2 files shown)


mirroring/mirroring.go
```diff
@@ -136,46 +137,88 @@ func VerifySignedTreeHead() error {
 	return nil
 }
 
-func ComputeNewRootHash(oldRootHash string, logProof *models.ConsistencyProof) (string, error) {
-	newRootHash, err := hex.DecodeString(oldRootHash)
+func VerifyLogConsistency(oldSize int64, oldRootHash string) error {
+	rekorClient, err := client.GetRekorClient(rekorServerURL)
```
SantiagoTorres, line 141: Hmm, how is rekorServerURL resolved here?

mirroring/mirroring.go
```diff
@@ -136,46 +137,88 @@ func VerifySignedTreeHead() error {
...
+		hash, err := hex.DecodeString(h)
 		if err != nil {
-			return "", err
+			return err
 		}
-		newRootHash = rfc6962.DefaultHasher.HashChildren(newRootHash, hash)
+		proofs[i] = hash
+	}
+
+	verifier := logverifier.New(rfc6962.DefaultHasher)
```
SantiagoTorres, line 175: I wonder if the hasher here can be loaded from a location in the rekor codebase later...

SantiagoTorres (approved):
Two minor nits, looks good to me  if they can be answered.

mirroring/mirroring.go
hojoung97: It is set as a constant on line 44 like this `const rekorServerURL = "https://api.sigstore.dev"`.
I was not sure if this is a good idea. Let me know if you have any other suggestion!

mirroring/mirroring.go
hojoung97: For a short answer, I think we can do that! But to understand better, will that be something like switching to different hasher or initializing with a custom/user-defined hasher?

mirroring/mirroring.go
SantiagoTorres: Hmm, for now let's leave it as is, but I'd assume we want this to be overriddable (for the sake of testing and such). Can you make a ticket about making the api url a configuration parameter/setting?

mirroring/mirroring.go
SantiagoTorres: Can you make a ticket about this too? maybe a question tag for it :)

SantiagoTorres:
Once you're done with the tickets I think we can go ahead and merge. Cheers!

Issue #29, opened 2021-09-21
Configurable rekorServerURL in mirroring.go


hojoung97:
**Description**

Currently, `rekorServerURL` is fixed as `https://api.sigstore.dev` in `mirroring.go`.
This should be a configurable parameter passed into most functions in `mirroring.go` when initializing a Rekor client instance. 

Refer to #28

PR #31, opened 2021-09-21
Make rekor client instance as an input parameter

hojoung97:
Signed-off-by: Hojoung Jang <jang88@purdue.edu>

#### Summary

Make necessary refactoring to implement main executable for mirroring job
- Removed rekor client instantiation in all of functions and make it as input parameter
  - This will allow us to resolve any rekor server URL of choice upon rekor client instantiation
- `VerifySignedTreeHead()` -> `VerifySignedTreeHead(logInfo *models.LogInfo, pubkey string)`
  - Takes in `logInfo` containing signed tree head information and corresponding server's public key
  - It made more sense to verify given signed tree head rather than doing querying inside the function and verifying
- `VerifyLogConsistency()` returns the new tree size and root hash upon successful consistency verification
  - Made more sense to return this information so that the old tree data gets updated

#### Ticket Link

Fixes https://github.com/sigstore/rekor-monitor/issues/29

#### Release Note

```release-note

```

### DIFF
--- a/mirroring/mirroring.go
+++ b/mirroring/mirroring.go
@@ -41,8 +41,6 @@ import (
 	"github.com/spf13/viper"
 )
 
-const rekorServerURL = "https://api.sigstore.dev"
-
 type getCmdOutput struct {
 	Body           types.EntryImpl
 	LogIndex       int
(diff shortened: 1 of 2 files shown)


SantiagoTorres (commented):
LGTM!

I think we should start documenting this a little bit more clearly. In particular the retval of `VerifyLogConsistency` may be confusing later down the line.

SantiagoTorres (approved):

PR #37, opened 2021-10-25
Set up gRPC service

hojoung97:
#### Summary
Add gRPC service - implements a RPC called `GetLastSnapshot` that returns latest log info that has been verified by the monitor mirror job
- `internal/monitorGRPC` contains
   - proto files
   - generated gRPC code
   - gRPC server implementation (`monitorGRPCServer.go`)
-  `cmd/grpcServer` contains
   - gRPC server main executable

Other Changes:
- Move log info file reading from `cmd/mirroring/main.go` to `mirroring/utils.go`
- Add consistency check fail error to identify it from other errors (`mirroring/mirroring.go`)

#### Release Note

```release-note

```


cmd/grpcServer/main.go
```diff
@@ -0,0 +1,45 @@
...
+		port = val
+	}
+
+	lis, err := net.Listen("tcp", fmt.Sprintf(":%s", port))
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	grpcServer := grpc.NewServer()
+	monitorGRPC.RegisterMonitorServiceServer(grpcServer, monitorGRPC.NewServer())
```
A0su, line 40: Will we not need access to the newly instantiated `monitorGRPC.NewServer()` at some point in the future for the argument `srv MonitorServiceServer`?

cmd/grpcServer/Dockerfile
```diff
@@ -0,0 +1,34 @@
...
+# limitations under the License.
+
+# Apply multi-stage build
+FROM golang:alpine AS builder
+WORKDIR /builder
+
+# Copy go source files for build
+COPY go.mod ./
+COPY go.sum ./
+COPY cmd/grpcServer/main.go ./
```
A0su, line 23: Could coalesce three copies into one but I don't believe there's any benefit at the current moment

mirroring/utils.go
```diff
@@ -0,0 +1,54 @@
...
+
+	// Read line by line and get the last line
+	scanner := bufio.NewScanner(file)
+	line := ""
+	for scanner.Scan() {
+		line = scanner.Text()
+	}
+
+	// Each line is in the format of space-separeted info: "treeSize rootHash"
+	parsed := strings.Split(line, " ")
```
A0su, line 43: Do we want error handling if `Split()` returns an empty slice?

A0su (commented):
Few minor q's but rest looks great and build works properly on my machine

mirroring/utils.go
hojoung97: Yes please!

cmd/grpcServer/Dockerfile
hojoung97: Can you collapse into one? I would like to see how that's done :)

cmd/grpcServer/main.go
hojoung97: Yes, that would be better

cmd/grpcServer/Dockerfile
A0su: Yup! The way I usually have written it before is as `COPY go.mod go.sum cmd/grpcServer/main.go ./`.

haydentherapper:
Tracking in https://github.com/sigstore/rekor-monitor/issues/91

PR #40, opened 2022-07-07
Run periodically on GitHub Actions

imjasonh:
Signed-off-by: Jason Hall <imjasonh@gmail.com>

#### Summary

This adds a GitHub Actions workflow to build and run the monitor code on an hourly schedule. `logInfo.txt` is not persisted between runs, so each run is the "first" run in this case.

This also updates the `--interval` flag to take a duration string (e.g., `--interval=5m`, `--interval=10s`, `--interval=6h`) instead of an integer number of minutes.

It also fixes a few tiny Go nits around error value scoping and usage of `log.Print` and `log.Fatal`.

Run this way, if the periodic scheduled verification fails, an email will be sent to repo owners to alert them. We can also create a badge for this on the repo page.

#### Release Note

Added a `--once` flag to run verification once then exit. `--interval` is updated to take a Go duration string (e.g., `5m`, `10s`, `6h`), and more errors result in failure instead of just logging.

#### Documentation

N/A

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -18,7 +18,7 @@ name: Check
 on:
   workflow_dispatch:
   schedule:
-    cron: '0 * * * *' # every hour
+  - cron: '0 * * * *' # every hour
 
 jobs:
   check:
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -24,8 +24,8 @@ jobs:
   check:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
-      - uses: actions/setup-go@v2
+      - uses: actions/checkout@2541b1294d2704b0964813337f33b291d3f8596b # v2.4.0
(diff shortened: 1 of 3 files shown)


.github/workflows/check.yml
```diff
@@ -0,0 +1,31 @@
...
+  workflow_dispatch:
+  schedule:
+  - cron: '0 * * * *' # every hour
+
+jobs:
+  check:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-go@v2
```
bobcallaway, line 28: nit: can you pin these to hashes instead of tags?

.github/workflows/check.yml
```diff
@@ -0,0 +1,31 @@
...
+        with:
+          go-version: 1.18
```
bobcallaway, line 30: you might want to bump `go.mod` to match or downlevel this

bobcallaway (commented):
lgtm other than the two nits

bobcallaway (approved):

imjasonh:
Anything else this needs to be able to merge it?

PR #42, opened 2022-08-24
Check consistency between persisted snapshots

### LINKED ISSUE #41: Updating the GH cronjob to persist state
**Description**

https://github.com/sigstore/rekor-monitor/pull/40 added consistency checking periodically. However, this doesn't check consistency as expected. After each invocation, a log file is written to disk with the previous checkpoint. Consistency is checked between the previous checkpoint and the current tree head. https://github.com/sigstore/rekor-monitor/blob/main/mirroring/mirroring.go#L122

If this is invoked on GH without state persisted between each run, what happens is:
1) We fetch the latest tree head - https://github.com/sigstore/rekor-monitor/blob/main/cmd/mirroring/main.go#L107
2) We verify consistency - https://github.com/sigstore/rekor-monitor/blob/main/cmd/mirroring/main.go#L146. This fetches the latest tree head, which either might not have changed, or maybe a couple of entries have been appended.

Without persisting the latest checkpoint, we're effectively just checking an inclusion proof on the current tree head, or a consistency proof between two checkpoints that are very near one another.

haydentherapper:
A previous PR added an hourly cronjob to check consistency. This didn't
check consistency as expected. After each invocation, a log file is
written to disk with the previous checkpoint. Consistency is checked
between the previous and current checkpoints. Without persistence, the
job would check consistency between the same checkpoints.

To implement persistence, we use a GitHub Action that downloads
previously uploaded artifacts. Persistence between workflows isn't
natively supported.

This also changes the persisted format to be a signed checkpoint rather
than just the size and root hash, which means we can verify it.

This also switches the default URL to be rekor.sigstore.dev, as api was
deprecated.

Fixes #41


haydentherapper:
cc @asraa @SantiagoTorres @bobcallaway

haydentherapper:
Note there are some TODOs, I'll be creating issues for everything.

cmd/mirroring/main.go
```diff
@@ -54,17 +55,56 @@ func readLogInfo(treeSize *int64, root *string) error {
...
+	}
+
+	// open file again to overwrite
+	file, err = os.OpenFile(logInfoFile, os.O_RDWR|os.O_TRUNC, 0666)
+	if err != nil {
+		return err
+	}
+	defer file.Close()
+
+	// keep all but the oldest checkpoint
```
asraa, line 101: Instead of a scanner, use a `bufio.NewReader` to `ReadBytes` for 100 lines, and maintain the end position, then just truncate the file to the calculated length.

something like
```
r := bufio.NewReader(file)
endPos := 0
for numLines := 0;  numLines < 100; numLines++ {
    line, err = r.ReadBytes('\n')
    if err != nil { return err }
    endPos += len(line)
}
if err := os.Truncate(logInfoFile, endPos); err != nil {
   return err
}
```

cmd/mirroring/main.go
```diff
@@ -79,85 +119,104 @@ func main() {
...
+		// Replace newlines to flatten checkpoint to single line
+		if _, err := file.WriteString(fmt.Sprintf("%s\n", strings.Replace(string(s), "\n", "\\n", -1))); err != nil {
+			log.Fatalf("failed to write to file: %v", err)
 		}
 	}
 
 	for {
 		// Check for root hash consistency
-		newTreeSize, newRoot, err := mirroring.VerifyLogConsistency(rekorClient, treeSize, root)
+		newSTH, err := mirroring.VerifyLogConsistency(rekorClient, int64(sth.Size), sth.Hash)
```
asraa, line 186: meta comment: the overall pattern in the first STH case is very strange: get an STH, then verifylogconsistency with another newSTH retrieved seconds later?

if this is the first, then why not JUST get an STH, verify it, write it, and exit. if there's existing, load the old state, and verify consistency with a new one and write.

so something like:
1. Get new STH & verify
2. if old STH if exists, load and verify log consistency
3. Write new STH

This avoid the two `GetLogInfo`s that happen on first case.

mirroring/mirroring.go
```diff
@@ -119,42 +114,41 @@ func VerifySignedTreeHead(logInfo *models.LogInfo, pubkey string) error {
...
 		if err != nil {
-			return 0, "", err
+			return nil, err
 		}
 		proofs[i] = hash
 	}
 
 	verifier := logverifier.New(rfc6962.DefaultHasher)
-	err = verifier.VerifyConsistencyProof(oldSize, *logInfo.TreeSize, oldRoot, newRoot, proofs)
+	err = verifier.VerifyConsistencyProof(oldSize, *logInfo.TreeSize, oldRootHash, newRoot, proofs)
```
asraa, line 143: note: this is deprecated https://github.com/sigstore/rekor/pull/807

.github/workflows/check.yml
```diff
@@ -28,4 +28,24 @@ jobs:
       - uses: actions/setup-go@b22fbbc2921299758641fab08929b4ac52b32923 # v2.2.0
         with:
           go-version: 1.16
+      - name: Download artifact
+        uses: dawidd6/action-download-artifact@ea71f332a90fd52416b1629b933dcf7e0ccd421d # v2.22.0
+        with:
+          workflow: check.yml
+          run_id: ${{ github.event.workflow_run.id }}
```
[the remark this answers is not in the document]
haydentherapper: Removed - We found that there's a bug where `run_id` isn't evaluated, so this was working accidentally.

.github/workflows/check.yml
```diff
@@ -28,4 +28,24 @@ jobs:
...
+          name: checkpoint
+        # Skip on first run since there will be no checkpoint
+        continue-on-error: true
+      - name: Log current checkpoints
+        run: cat logInfo.txt
```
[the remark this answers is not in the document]
haydentherapper: Added.

.github/workflows/check.yml
```diff
@@ -13,12 +13,12 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-name: Check
+name: Consistency proof
```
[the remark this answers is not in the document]
haydentherapper: Added

mirroring/mirroring.go
haydentherapper: Updated to the latest rekor and go 1.19, and fixed

cmd/mirroring/main.go
haydentherapper: The problem is we need to drop the first X lines from the file, not the last. Truncate drops only the end, instead of the beginning.

cmd/mirroring/main.go
haydentherapper: Yep, very much agree! I was getting complacent with the existing code. Lemme add a TODO to clean that up in the next PR to avoid this PR being too messy.

cmd/mirroring/main.go
asraa: That's even easier! But that's not what this function is doing. The latest checkpoint is at the start, right? Trunacting would drop the oldest ones.

haydentherapper: I started rewriting this but realized we still need a scanner to confirm how many lines are in the file. You could do something like counting `\n`, but I don't see much benefit to that. Also planning to remove this code once we start writing history to git.

haydentherapper: I am updating this to correctly persist only the last 100, not just the single oldest.

asraa (approved):

PR #61, opened 2022-10-05
Fix broken tests, use Rekor verification functions

haydentherapper:
This fixes a broken test due to sharding. We also need to update the tests to not make network calls.


cmd/mirroring/main.go
```diff
@@ -189,7 +195,7 @@ func main() {
 
 	for {
 		// Check for root hash consistency
-		newSTH, err := mirroring.VerifyLogConsistency(rekorClient, int64(sth.Size), sth.Hash)
+		newSTH, err := verify.VerifyCurrentCheckpoint(context.Background(), rekorClient, verifier, sth)
```
asraa, line 198: This does the `sth.Verify`, so not sure if you need this above: if you remove it above, add a comment here saying it verifies the current checkpoint and verifies consistency against a newly fetched Checkpoint

mirroring/mirroring.go
```diff
@@ -76,115 +73,18 @@ func GetLogInfo(rekorClient *gclient.Rekor) (*models.LogInfo, error) {
...
-		return nil, err
-	}
-	sth := util.SignedCheckpoint{}
-	if err := sth.UnmarshalText([]byte(*logInfo.SignedTreeHead)); err != nil {
 		return nil, err
 	}
-	return &sth, nil
-}
-
-func VerifyLogInclusion(rekorClient *gclient.Rekor, entryUUID string) error {
```
asraa, line 153: was this just not used?

haydentherapper: Yep, this was not used. There's more functions in here that aren't used, but they don't have equivalents in rekor/pkg/verify, so I left them.

cmd/mirroring/main.go
haydentherapper: Whoops, I merged in the other PR which merged this commit. I'll update this!

PR #101, opened 2022-11-09
Add reusable workflow and identity checking, file issue on failure

haydentherapper:
This is the start to supporting identity checking in the monitor.

I also moved the workflow into a reusable workflow, so that going forward, users won't need to fork the repository, they can simply adopt the reusable worklow.

This PR got a little big, so you'll notice a lot of TODOs that I'll be filing issues for. For now, identity checking supports rekord and hashedrekord entries, which should cover most entries. There's commented out code from when I started supporting other entry types, but tests were hard to write :)

This also added error reporting (thanks for the script Asra!), where an issue will be filed if the monitor fails.


haydentherapper:
fyi @di, identity checking via GHA is coming!

.github/workflows/reusable_consistency_proof.yaml
```diff
@@ -0,0 +1,72 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@93ea575cb5d8a053eaa0ac8fa3b40d7e05a33cc8 # v2.4.0
+      - uses: actions/setup-go@c4a742cab115ed795e34d4513e2cf7d472deb55f # v2.2.0
+        with:
+          go-version: 1.19
+      - name: Download artifact
+        uses: dawidd6/action-download-artifact@b12b127cf24433d14b4f93cee62f5465076ba82a # v2.22.0
+        with:
+          workflow: check.yml
```
asraa, line 41: how will this work inside the reusable workflow? should you detect the name of the caller workflow?

.github/workflows/reusable_consistency_proof.yaml
```diff
@@ -0,0 +1,72 @@
...
+      - name: Upload checkpoint
+        uses: actions/upload-artifact@83fd05a356d7e2593de66fc9913b3002723633cb # v3.1.0
+        with:
+          name: ${{ env.UPLOADED_LOG_NAME }}
+          path: ${{ env.LOG_FILE }}
+          retention-days: 1
+      - name: Log new checkpoints
+        run: cat ${{ env.LOG_FILE }}
+
+  if-failed:
```
asraa, line 59: this could be one of the parameters (file issue/not)

.github/workflows/reusable_consistency_proof.yaml
```diff
@@ -0,0 +1,72 @@
...
+    runs-on: ubuntu-latest
+    needs: [check]
+    permissions:
+      issues: 'write'
+    env:
+      GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      ISSUE_REPOSITORY: sigstore/
```
asraa, line 66: right now i think this will fail without the repo name

pkg/rekor/client.go
```diff
@@ -0,0 +1,38 @@
...
+	"github.com/sigstore/rekor/pkg/generated/client"
+	"github.com/sigstore/rekor/pkg/generated/client/entries"
+	"github.com/sigstore/rekor/pkg/generated/models"
+)
+
+// GetEntriesByIndexRange fetches all entries by log index, from [start, end]
+func GetEntriesByIndexRange(ctx context.Context, rekorClient *client.Rekor, start, end int) ([]models.LogEntry, error) {
+	var logEntries []models.LogEntry
+	for i := start; i <= end; i++ {
+		params := entries.NewGetLogEntryByIndexParamsWithContext(ctx)
```
asraa, line 29: I think you can use the entries/retrieve endpoint with a list of logindexes (max 50)

cmd/mirroring/main.go
```diff
@@ -120,13 +123,28 @@ func main() {
 	interval := flag.Duration("interval", 5*time.Minute, "Length of interval between each periodical consistency check")
 	logInfoFile := flag.String("file", logInfoFileName, "Name of the file containing initial merkle tree information")
 	once := flag.Bool("once", false, "Perform consistency check once and exit")
+	identities := flag.String("identities", "{}", "List of identities and issuers to monitor, as JSON")
```
asraa, line 126: just curious about this, but what about removing JSON and just requesting it on a line by line separated

SUBJECTA ISSUERA.0 ISSUERA.1
SUBJECTB ISSUERB
SUBJECTC

cmd/mirroring/main.go
```diff
@@ -237,6 +259,36 @@ func main() {
...
+				for _, idEntry := range idEntries {
+					log.Printf("Found subject %s, issuer %s at log index %d, uuid %s\n",
+						idEntry.Subject, idEntry.Issuer, idEntry.Index, idEntry.UUID)
+
+					idFile, err := os.OpenFile(*identitiesFoundFile, os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0644)
+					if err != nil {
+						idFile.Close()
+						log.Fatalf("failed to open identities file: %v", err)
+					}
+					if _, err := idFile.WriteString(fmt.Sprintf("%s %s %d %s\n", idEntry.Subject, idEntry.Issuer, idEntry.Index, idEntry.UUID)); err != nil {
```
asraa, line 283: it's curious you chose JSON as the input format but chose the line separate output as format! consider choosing one or the other :)

cmd/mirroring/main.go
```diff
@@ -237,6 +259,36 @@ func main() {
...
+						idFile.Close()
+						log.Fatalf("failed to write to file: %v", err)
+					}
+					idFile.Close()
```
asraa, line 287: make this a defer statement instead of putting before every exit

pkg/rekor/identity.go
```diff
@@ -0,0 +1,210 @@
...
+				// TODO: Add support for TUF?
+				continue
+			}
+
+			// TODO: Support regular expressions using co.Identities
+			// TODO: Support GitHub CI claims
+
+			var checks []*cosign.CheckOpts
+			for _, ids := range identities.Identities {
+				// skip empty subjects
```
asraa, line 83: should this be enforced on parsing the input file?

pkg/rekor/identity.go
```diff
@@ -0,0 +1,210 @@
...
+					for _, iss := range ids.Issuers {
+						checks = append(checks, &cosign.CheckOpts{CertIdentity: ids.Subject, CertOidcIssuer: iss})
+					}
+				}
+			}
+
+			for _, co := range checks {
+				err = cosign.CheckCertificatePolicy(certs[0], co)
+				// certificate matched policy
+				if err == nil {
```
asraa, line 101: you can put this inline 
```
if err := cosign.Check; err == nil {}
```

cmd/mirroring/main.go
haydentherapper: Linter complains about log.Fatalf occurring before the file closes unfortunately - The right fix is to not call log.fatalf and fail gracefully. I'll file an issue

.github/workflows/check.yml
```diff
@@ -23,35 +23,8 @@ on:
...
-        uses: actions/upload-artifact@83fd05a356d7e2593de66fc9913b3002723633cb # v3.1.0
-        with:
-          name: ${{ env.UPLOADED_LOG_NAME }}
-          path: ${{ env.LOG_FILE }}
-          retention-days: 1
-      - name: Log new checkpoints
-        run: cat ${{ env.LOG_FILE }}
+  run_consistency_proof:
+    permissions:
+      contents: 'read'
```
asraa, line 29: do you need artifact write perms?

.github/workflows/reusable_consistency_proof.yaml
asraa: you should should be able to detect the name of the caller workflow (which is what's needed here)

.github/workflows/reusable_consistency_proof.yaml
```diff
@@ -0,0 +1,72 @@
...
+
+env:
+  UPLOADED_LOG_NAME: checkpoint
+  LOG_FILE: checkpoint_log.txt
+
+jobs:
+  check:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@93ea575cb5d8a053eaa0ac8fa3b40d7e05a33cc8 # v2.4.0
```
haydentherapper, line 34: @asraa says you need to hardcode the reusable workflow repository, else it checks out the caller workflow repository 

For repo ref: You need to use OIDC token to find the reference for the workflow that was specified in the caller. Ask Asra for the GHA for doing this

.github/workflows/reusable_consistency_proof.yaml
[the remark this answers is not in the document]
haydentherapper: Fetch repository name from context

.github/workflows/check.yml
```diff
@@ -23,35 +23,8 @@ on:
 permissions:
```
haydentherapper, line 23: permissions: read-all

.github/workflows/check.yml
haydentherapper, line 29: Must have "issues: write" if you allow issue posting

haydentherapper: I think no?

.github/workflows/check.yml
haydentherapper, line 23: Think about concurrency:

* Either tell people just don't
* For SLSA, put a random number at the end of the artifact, but that will complicate how we fetch the workflow artifact.

.github/workflows/scripts/report_failure.sh
```diff
@@ -0,0 +1,53 @@
...
+EOF
+}
+
+THIS_FILE=$(this_file)
+create_issue_body
+
+ISSUE_ID=$(gh -R "$ISSUE_REPOSITORY" issue list --label "bug" --state open -S "$THIS_FILE" --json number | jq '.[0]' | jq -r '.number' | jq 'select (.!=null)')
+
+if [[ -z "$ISSUE_ID" ]]; then
+    # Replace `-`` by ` `, remove the last 4 characters `.yml`. Expected: "snapshot timestamp".
```
haydentherapper, line 48: Fix this - Hardcode Rekor monitor

.github/workflows/reusable_consistency_proof.yaml
asraa: https://github.com/slsa-framework/slsa-github-generator/tree/main/.github/actions/detect-workflow

.github/workflows/check.yml
haydentherapper: Added a warning

pkg/rekor/client.go
haydentherapper: It seems to be 10 based on the openapi spec, I thought it was 50 too...

cmd/mirroring/main.go
haydentherapper: picked a more basic format

we'll probably need to revisit it for keys too, or maybe just ask for key fingerprints

cmd/mirroring/main.go
haydentherapper: i like it! went with that

haydentherapper:
Closing in favor of https://github.com/sigstore/rekor-monitor/pull/131/ - All comments have been addressed there

Issue #48, opened 2022-08-27
GHA: Create reusable workflow


haydentherapper:
**Description**

Once https://github.com/sigstore/rekor-monitor/pull/42 is merged, we should convert the workflow into a reusable workflow so it can be customized by others and not run by default for a forked repository.

PR #129, opened 2023-01-05
Add reusable workflow for monitoring

haydentherapper:
Fixes https://github.com/sigstore/rekor-monitor/issues/48

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -1,4 +1,3 @@
-#
 # Copyright 2022 The Sigstore Authors.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
@@ -20,38 +19,14 @@ on:
     - cron: '0 * * * *' # every hour
   workflow_dispatch:
 
-permissions:
-  contents: read
-
-env:
-  UPLOADED_LOG_NAME: checkpoint
-  LOG_FILE: checkpoint_log.txt
+permissions: read-all
 
 jobs:
-  check:
-    runs-on: ubuntu-latest
-    steps:
-      - uses: actions/checkout@755da8c3cf115ac066823e79a1e1788f8940201b # v2.4.0
-      - uses: actions/setup-go@6edd4406fa81c3da01a34fa6f6343087c207a568 # v2.2.0
-        with:
-          go-version: 1.19
-      - name: Download artifact
-        uses: dawidd6/action-download-artifact@e6e25ac3a2b93187502a8be1ef9e9603afc34925 # v2.22.0
-        with:
-          workflow: check.yml
-          name: ${{ env.UPLOADED_LOG_NAME }}
-        # Skip on first run since there will be no checkpoint
-        continue-on-error: true
-      - name: Log current checkpoints
-        run: cat ${{ env.LOG_FILE }}
-        # Skip on first run
-        continue-on-error: true
-      - run: go run ./cmd/mirroring --file ${{ env.LOG_FILE }} --once
-      - name: Upload checkpoint
-        uses: actions/upload-artifact@83fd05a356d7e2593de66fc9913b3002723633cb # v3.1.0
-        with:
-          name: ${{ env.UPLOADED_LOG_NAME }}
-          path: ${{ env.LOG_FILE }}
-          retention-days: 1
-      - name: Log new checkpoints
-        run: cat ${{ env.LOG_FILE }}
+  run_consistency_proof:
+    permissions:
+      contents: read
+      issues: write
+      id-token: write # Needed to detect the current reusable repository and ref.
+    uses: sigstore/rekor-monitor/.github/workflows/reusable_monitoring.yaml@main
+    with:
+      file_issue: true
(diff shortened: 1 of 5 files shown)


.github/workflows/check.yml
```diff
@@ -13,45 +12,22 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# NOTE: This GHA should not be run concurrently.
```
haydentherapper, line 15: Note to self: Move this to the reusable workflow

.github/workflows/check.yml
```diff
@@ -20,38 +19,14 @@ on:
...
-          name: ${{ env.UPLOADED_LOG_NAME }}
-          path: ${{ env.LOG_FILE }}
-          retention-days: 1
-      - name: Log new checkpoints
-        run: cat ${{ env.LOG_FILE }}
+  run_consistency_proof:
+    permissions:
+      contents: read
+      issues: write
+      id-token: write
```
asraa, line 29: Hmmm I don't *think* you need `id-token: write` -- no OIDC access going on! (I think)

.github/workflows/reusable_monitoring.yml
```diff
@@ -0,0 +1,93 @@
...
+      - name: Upload checkpoint
+        uses: actions/upload-artifact@83fd05a356d7e2593de66fc9913b3002723633cb # v3.1.1
+        with:
+          name: ${{ env.UPLOADED_LOG_NAME }}
+          path: ${{ env.LOG_FILE }}
+          retention-days: 1
+      - name: Log new checkpoints
+        run: cat ${{ env.LOG_FILE }}
+
+  if-failed:
```
asraa, line 77: I think you may also need an `if-succeed` that can run to close any existing issues that had detected a failure (see  slsa-framework/example-package for a script)

.github/workflows/reusable_monitoring.yml
```diff
@@ -0,0 +1,93 @@
...
+    runs-on: ubuntu-latest
+    needs: [monitor]
+    permissions:
+      issues: 'write'
+    env:
+      GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      ISSUE_REPOSITORY: ${{ steps.detect-workflow.outputs.repository }}
```
asraa, line 84: this would be sigstore/rekor-monitor. an issue should be filed in THEIR repo right? if so, I think `$${ github.repository }}` will work

haydentherapper: changed

.github/workflows/reusable_monitoring.yml
haydentherapper: added

haydentherapper:
Shown working in haydentherapper/rekor-log-monitor

* It picks up a previous successful run in https://github.com/haydentherapper/rekor-log-monitor/actions/runs/3850712338
* It closes a failed issue on success - https://github.com/haydentherapper/rekor-log-monitor/actions/runs/3850684787/jobs/6561137869
* It logs an issue on failure - https://github.com/haydentherapper/rekor-log-monitor/actions/runs/3850184576/jobs/6560055103 (note this run succeeded, but a bug forced it to log an issue)

.github/workflows/reusable_monitoring.yml
```diff
@@ -0,0 +1,112 @@
...
+        run: cat ${{ env.LOG_FILE }}
+        # Skip on first run
+        continue-on-error: true
+      - run: go run ./cmd/mirroring --file ${{ env.LOG_FILE }} --once
+      - name: Upload checkpoint
+        uses: actions/upload-artifact@83fd05a356d7e2593de66fc9913b3002723633cb # v3.1.1
+        with:
+          name: ${{ env.UPLOADED_LOG_NAME }}
+          path: ${{ env.LOG_FILE }}
+          retention-days: 1
```
asraa, line 74: Very minor question here: what happens if someone runs their monitor weekly? In that case there would be no consistency measured across because the checkpoints would disappear.

Do you want to document that it's intended to be run on a daily schedule (at least)?

It seems hairier to detect the callers schedule. You can via `github.event.schedule`, but I can imagine you'd need to do some parsing and so on on the cron and make a decision for what the minimum time should be

asraa (approved):
LGTM minus the comment! Nice work with all the testing, I know how annoying it can get :)

.github/workflows/reusable_monitoring.yml
haydentherapper: Excellent point, I'm going to just provide an input to allow the caller to set retention and document this as a limitation. I've increased it to 14 days also - The artifacts are tiny, so I'm not worried about storage limits.

haydentherapper:
@asraa ready for another review

https://github.com/haydentherapper/rekor-log-monitor/actions/runs/3857676546 for another successful run after this change

.github/workflows/reusable_monitoring.yml
asraa: perfect!

asraa (approved):
Great!!! Love how it turned out.

Issue #52, opened 2022-08-27
Consistency: Change order of operations


haydentherapper:
**Description**

As suggested in #42, we should:
* Fetch latest STH and verify
* If old STH is present, very consistency proof
* Write new STH to log

This avoids the awkward behavior for the first STH fetch, which fetches an STH, logs it, then immediately fetches another and logs it.

PR #131, opened 2023-01-06
Add support for monitoring identities

haydentherapper:


haydentherapper:
@asraa I haven't tested this on GHA, but code should be ready for review

Tested locally

cmd/mirroring/main.go
```diff
@@ -237,6 +285,36 @@ func main() {
...
+				log.Fatalf("error getting entries by index range: %v", err)
+			}
+			idEntries, err := rekor.MatchedIndices(entries, ids)
+			if err != nil {
+				log.Fatalf("error finding log indices: %v", err)
+			}
+
+			if len(idEntries) > 0 {
+				for _, idEntry := range idEntries {
+					log.Printf("Found subject %s, issuer %s at log index %d, uuid %s\n",
```
asraa, line 301: log to StdErr maybe?

cmd/mirroring/main.go
```diff
@@ -237,6 +285,36 @@ func main() {
...
+						idEntry.Subject, idEntry.Issuer, idEntry.Index, idEntry.UUID)
+
+					idFile, err := os.OpenFile(*identitiesFoundFile, os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0644)
+					if err != nil {
+						idFile.Close()
```
asraa, line 306: use a `defer idFile.Close()` after opening was successful so you can remove the previous calls.

pkg/rekor/identity.go
```diff
@@ -0,0 +1,164 @@
...
+}
+
+type Identity struct {
+	Subject string   `json:"subject"`
+	Issuers []string `json:"issuers,omitempty"`
+}
+
+// MatchedIndices returns a list of log indices that contain the requested identities.
+func MatchedIndices(logEntries []models.LogEntry, identities Identities) ([]IdentityEntry, error) {
+	if err := verifyIdentities(identities); err != nil {
```
asraa, line 52: shouldn't this check occur once at input ingestion?

cmd/mirroring/main.go
```diff
@@ -206,6 +246,14 @@ func main() {
 			log.Printf("Root hash consistency verified - Tree Size: %d Root Hash: %s\n", newSTH.Size, hex.EncodeToString(newSTH.Hash))
 		}
 
+		// Get log size of inactive shards
```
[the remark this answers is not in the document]
haydentherapper: yea, this ordering issue has come up before. i have a TODO on line 177 to refactor this so we only write it once. I'll handle that in the next PR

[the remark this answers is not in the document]
haydentherapper: https://github.com/sigstore/rekor-monitor/issues/52

pkg/rekor/identity.go
asraa: Hmmm agree, but on the other hand if the identities are badly formed as inputs, you probably don't want to do all the STH work before alerting on it. But not a big deal!

pkg/rekor/identity.go
```diff
@@ -0,0 +1,164 @@
...
+	}
+
+	var matchedIndices []IdentityEntry
+
+	for _, entries := range logEntries {
+		for uuid, entry := range entries {
+			entry := entry
+
+			certs, err := extractCertificates(&entry)
+			if err != nil {
```
asraa, line 63: Do you want to add a log statement here like `Skipping entry xyz`, or do you want to differ between the errors that are unexpected (unmarshalling the entry type, decoding string) and the ones that are possible (other entry types)?

asraa (approved):
Nice! Nothing blocking now.

pkg/rekor/identity.go
haydentherapper: Makes sense, I'll add the check! It's a subset, since we are ok with empty identities

pkg/rekor/identity.go
```diff
@@ -0,0 +1,164 @@
...
+	}
+
+	eimpl, err := types.CreateVersionedEntry(pe)
+	if err != nil {
+		return nil, err
+	}
+
+	var publicKeyB64 []byte
+	switch e := eimpl.(type) {
+	case *hashedrekord_v001.V001Entry:
```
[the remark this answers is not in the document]
haydentherapper: Now there is! https://github.com/sigstore/rekor-monitor/issues/136

Just blocked by the rekor verifier PR

pkg/rekor/identity.go
haydentherapper: I'm thinking to skip a log statement because the plan is to add support for all entry types and all key formats. It's just a limitation of the current tool. Also, given the number of entries we're ingesting, that'd be a lot of output.

pkg/rekor/identity.go
haydentherapper: Actually just checked, it shouldn't be possible to pass an identity without a subject because empty lines are skipped over.

pkg/rekor/identity.go
asraa: Ah, I guess with the verifier identities it should be easier to discern between expected and unexpected

PR #143, opened 2023-01-18
Update GHA to get notifications

haydentherapper:
I'm hoping this will change update who gets notified on action failure. https://docs.github.com/en/actions/monitoring-and-troubleshooting-workflows/notifications-for-workflow-runs says that only updating the cron syntax or enabling the workflow will change it, but given the last update to just change yaml->yml, I'm hoping this fixes notifications. 

Signed-off-by: Hayden B <hblauzvern@google.com>

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -26,7 +26,7 @@ jobs:
     permissions:
       contents: read
       issues: write
-      id-token: write # Needed to detect the current reusable repository and ref.
+      id-token: write # Needed to detect the current reusable repository and ref
     uses: sigstore/rekor-monitor/.github/workflows/reusable_monitoring.yml@main
     with:
       file_issue: true


asraa (commented):
what is the change here besides the commenting update?

haydentherapper:
I'm seeing if this changes who the "owner" of the GHA is. Right now, Bob is getting emails whenever they fail. I'd prefer I get the emails. This isn't a great setup regardless, but there isn't a way to configure a group of emails for GHA failures. I need to file a FR

asraa (approved):
OH got it. 

> Right now, Bob is getting emails whenever they fail. 

Yeah, I think it's the workflow author by default. That being said, I assume you get the file_issue emails?

haydentherapper:
I do see the file issue emails, yea!

PR #190, opened 2023-04-12
Update Cosign and sigstore/sigstore deps

haydentherapper:


haydentherapper:
nvm, thought I needed to update both at the same time. looks like a docker vuln report that is blocking this

Issue #155, opened 2023-02-11
Replace time.Sleep in main loop with ticket


haydentherapper:
**Description**

https://gobyexample.com/tickers - They're more idiomatic and non-blocking

PR #245, opened 2023-07-11
replaced time.Sleep in main loop with ticker

gntouts:
There is no change in functionality, just replaced time.Sleep in main loop with ticker. 

Changes:

- Replace time.Sleep in main loop with ticker

Fixes sigstore/rekor-monitor#155

### DIFF
--- a/cmd/verifier/main.go
+++ b/cmd/verifier/main.go
@@ -29,8 +29,10 @@ import (
 	"github.com/sigstore/rekor-monitor/pkg/rekor"
 	file "github.com/sigstore/rekor-monitor/pkg/util"
 	"github.com/sigstore/rekor/pkg/client"
+	gclient "github.com/sigstore/rekor/pkg/generated/client"
 	"github.com/sigstore/rekor/pkg/util"
 	"github.com/sigstore/rekor/pkg/verify"
+	"github.com/sigstore/sigstore/pkg/signature"
 )
 
 // Default values for monitoring job parameters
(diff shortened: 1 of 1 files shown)


cmd/verifier/main.go
```diff
@@ -101,7 +101,8 @@ func main() {
 	// 1. Fetch latest checkpoint and verify
 	// 2. If old checkpoint is present, verify consistency proof
 	// 3. Write latest checkpoint to file
-	for {
+	ticker := time.NewTicker(*interval)
```
haydentherapper, line 104: Can you add `defer ticket.Stop()` to clean up resources?

haydentherapper (changes requested):
Thanks so much for the PR! Just one nit

cmd/verifier/main.go
gntouts: Sure!

haydentherapper:
The linter failed because unfortunately we currently have some bad design in the main loop where we call `log.Fatal` in a lot of places. We need to do some refactoring at some point, but for now, we can explicitly call `ticker.Stop()` before calling `log.Fatal`, or try to do some refactoring.

gntouts:
Do you have any suggestions for refactoring? My initial thoughts would be to encapsulate a part of the main function within a separate function. This wrapper function should return an error containing the precise message that would be printed by log.Fatalf(). That way, the `defer ticker.Stop()` statement will run, while still allowing the use of log.Fatalf() in the main function to keep the user-facing messages unchanged.

haydentherapper:
That sounds great. Over time, I'd like to split up the main loop into smaller functions, but for now, wrapping it would be good.

gntouts:
I wrapped the loop in a separate function and modified the ticker loop to get an immediate fisrt tick. Let me know what you think.

PR #358, opened 2024-03-11
Bump go version to 1.22, bump linter

haydentherapper:

### DIFF
--- a/.github/workflows/verify.yml
+++ b/.github/workflows/verify.yml
@@ -48,7 +48,7 @@ jobs:
           check-latest: true
 
       - name: golangci-lint
-        uses: golangci/golangci-lint-action@3cfe3a4abbb849e10058ce4af15d205b6da42804 # v3
+        uses: golangci/golangci-lint-action@3cfe3a4abbb849e10058ce4af15d205b6da42804 # v4.0.0
         with:
-          version: v1.56
+          version: v1.57
           args: --timeout=5m --verbose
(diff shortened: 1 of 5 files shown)


haydentherapper:
For some reason, codeql is downloading 2.16.3 rather than 2.16.4 which includes go 1.22. will hold off for now

cpanato (dismissed):
besides the codeql
lgtm

haydentherapper:
Nothing to do at this point, codeql is being pulled from the hosted tool cache, so we just need to wait for that to be updated.

haydentherapper:
Toolchain now contains 2.16.4 but Go 1.22 support is still missing even though the docs say it should be present. will wait til 2.16.5 i guess

haydentherapper:
@cpanato finally figured it out, I sent the go.mod go version to the specific version of Go because apparently go 1.22 is a development version. hopefully this is correct now

.github/workflows/verify.yml
```diff
@@ -50,5 +50,5 @@ jobs:
       - name: golangci-lint
         uses: golangci/golangci-lint-action@3cfe3a4abbb849e10058ce4af15d205b6da42804 # v3
         with:
-          version: v1.54
+          version: v1.56
```
cpanato, line 53: can you update to `v1.57` ?

.github/workflows/verify.yml
```diff
@@ -50,5 +50,5 @@ jobs:
       - name: golangci-lint
         uses: golangci/golangci-lint-action@3cfe3a4abbb849e10058ce4af15d205b6da42804 # v3
```
cpanato, line 51: as you are here can you update the version comment to the full semver as well?

cpanato (changes requested):
thanks! some small cr

.github/workflows/verify.yml
haydentherapper: done

cpanato: thanks!

cpanato (approved):

PR #391, opened 2024-05-11
Add support for OID extension monitoring via YAML config

### LINKED ISSUE #158: [feature] Support OID extension monitoring
**Description**

SLSA GitHub generators use Sigstore signing to sign releases. Trusted builders use their GH provided OIDC identity to sign. The source repository is contained inside OID extensions, and later with enhancements, the caller workflow job ref will be as well. If rekor-monitor supports OID extension monitoring, users can monitor their package builds.

JonathanOppenheimer:
#### Summary

Closes #158

This PR adds the ability to define acceptable values for specific OIDs via the YAML configuration, in the same manner that acceptable fingerprints are constrained. Additional test cases have been added to `identity_test.go`. 

Note that this pull request introduces some idiosyncracies into the contraints. At the moment (at least as far I understand / have seen) entries which match *ANY* constraint will be added to `matchedEntries []IdentityEntry` which seems somewhat questionable, rather than entries that match *ALL* constraints. For OID constraints, I have only added them if they matched all defined constraints. 

This is most likely due to my lack of knowledge of the project. 

#### Release Note

> * New features and improvements, including behavioural changes, UI changes and CLI changes

Added capability to define OID constraints through OID and acceptable value pairs defined via the YAML config file 

#### Documentation
This PR requires very minor updates to the documentation, noting that users can now define OIDs in their YAML files. Users will implement the new feature by including OID constraints in their YAML file if they so choose.

### DIFF
--- a/pkg/rekor/identity.go
+++ b/pkg/rekor/identity.go
@@ -57,6 +57,11 @@ type CertificateIdentity struct {
 	Issuers     []string `yaml:"issuers"`
 }
 
+type OIDconstraint struct {
+	OID        []int  `yaml:"oid"`
+	Constraint []byte `yaml:"constraint"`
+}
+
 // MonitoredValues holds a set of values to compare against a given entry
 type MonitoredValues struct {
 	// CertificateIdentities contains a list of subjects and issuers
@@ -72,6 +77,8 @@ type MonitoredValues struct {
 	// Subjects contains a list of subjects that are not specified in a
 	// certificate, such as a SSH key or PGP key email address
 	Subjects []string `yaml:"subjects"`
+	// OIDConstraints contains a list of OIDs and constraints to be checked
+	OIDConstraints []OIDconstraint `yaml:"oidConstraints"`
 }
 
 // IdentityEntry holds a certificate subject, issuer, and log entry metadata
@@ -127,22 +134,6 @@ func MatchedIndices(logEntries []models.LogEntry, mvs MonitoredValues) ([]Identi
 				}
 			}
 
-			for _, monitoredCertID := range mvs.CertificateIdentities {
-				for _, cert := range certs {
-					match, sub, iss, err := certMatchesPolicy(cert, monitoredCertID.CertSubject, monitoredCertID.Issuers)
-					if err != nil {
-						return nil, fmt.Errorf("error with policy matching for UUID %s at index %d: %w", uuid, *entry.LogIndex, err)
-					} else if match {
-						matchedEntries = append(matchedEntries, IdentityEntry{
-							CertSubject: sub,
-							Issuer:      iss,
-							Index:       *entry.LogIndex,
-							UUID:        uuid,
-						})
-					}
-				}
-			}
-
 			for _, monitoredSub := range mvs.Subjects {
 				regex, err := regexp.Compile(monitoredSub)
 				if err != nil {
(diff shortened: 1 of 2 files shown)


haydentherapper:
Thanks for getting this implemented! I'll take a look shortly.

haydentherapper:
> At the moment (at least as far I understand / have seen) entries which match ANY constraint will be added to matchedEntries []IdentityEntry which seems somewhat questionable, rather than entries that match ALL constraints.

This is working as intended, as the intent with the current basic policy is to specify any identities you want to monitor for, and if any are found, they are recorded. Do you have a suggestion on how to convey this better? Do you think there's a need for a more complex policy to match multiple constraints?

pkg/rekor/identity.go
```diff
@@ -313,3 +349,56 @@ func certMatchesPolicy(cert *x509.Certificate, expectedSub string, expectedIssue
...
+		return false, fmt.Errorf("OID %s exists but does not match the expected value", oid)
+	}
+
+	return true, nil
+}
+
+// Helper functions
+
+// compareValue compares two byte arrays for equality
+func compareValue(value []byte, target []byte) bool {
```
haydentherapper, line 382: You can use `bytes.Equal`.

pkg/rekor/identity.go
```diff
@@ -57,6 +57,11 @@ type CertificateIdentity struct {
 	Issuers     []string `yaml:"issuers"`
 }
 
+type OIDconstraint struct {
```
haydentherapper, line 60:
```suggestion
type OIDConstraint struct {
```

pkg/rekor/identity.go
```diff
@@ -158,6 +149,40 @@ func MatchedIndices(logEntries []models.LogEntry, mvs MonitoredValues) ([]Identi
...
+							Index: *entry.LogIndex,
+							UUID:  uuid,
+						}] = struct{}{}
+					}
+				}
+			}
+
+			// If any oidConstraints were set, ensure only matched entries that meet all the constraints are returned
+			if len(oidConstraintSet) != 0 {
+				matchedEntries = FilterEntries(matchedEntries, oidConstraintSet)
```
haydentherapper, line 184: I would like the two cert-related sets to be separate - you can monitor by certificate subject alternative name, or you can monitor by OID value. We shouldn't constraint the former based on the latter, otherwise you'd be required to always match by cert subject as well as OID value.

For a concrete example, I may want to monitor for all instances of my repository. I could match against [Source Repo URI](https://github.com/sigstore/fulcio/blob/main/docs/oid-info.md#13614157264112--source-repository-uri) without the need to match by cert subject.

pkg/rekor/identity.go
```diff
@@ -72,6 +77,8 @@ type MonitoredValues struct {
 	// Subjects contains a list of subjects that are not specified in a
 	// certificate, such as a SSH key or PGP key email address
 	Subjects []string `yaml:"subjects"`
+	// OIDConstraints contains a list of OIDs and constraints to be checked
+	OIDConstraints []OIDconstraint `yaml:"oidConstraints"`
```
haydentherapper, line 81: I think we should change the struct so that we can match against a list of extension values given a single OID. With the current structure, you would need to specify an OID key-value pair repeatedly to match against multiple instances. So `OIDConstraint` becomes something like:

```
	OID      asn1.ObjectIdentifier  `yaml:"oid"`
	Values []string`yaml:"values"`
```

I would suggest we start with the assumption that all OID values are strings and we support only the Fulcio OIDs.

Note that this wouldn't support a more complex policy that would match against multiple extension values for a given certificate. For example, maybe you want to match all certificates where an repo is X AND a subject is Y. I think this should be tackled separately. This is a useful feature but would require changing all policy checks. We should explore using a policy language too rather than reinventing this.

haydentherapper (changes requested):

haydentherapper:
Just wanted to check in, did you have any questions about these suggestions?

haydentherapper:
Closing as this is currently being worked on.

PR #440, opened 2024-08-22
set user-agent on requests

bobcallaway:


haydentherapper (approved):
Did you want to set this in the GHA as well?

bobcallaway:
> Did you want to set this in the GHA as well?

we could pass something like the run ID in as a command line arg to help correlate, wdyt?

haydentherapper (approved):

PR #445, opened 2024-09-04
implement support for OID monitoring

linus-sun:
#### Summary

As documented [here](https://docs.google.com/document/d/1YcC31YmnCSwL1GRbC7LSor-YY6qhk3ynMhN1UNT-Wyw/edit?pli=1&resourcekey=0-OoutF_Qh1Q5q0p4ZPYXncA&tab=t.0#heading=h.xgjl2srtytjt), this PR aims to implement support for monitoring OID extension values in rekor-monitor. Users can input OID extension fields and associated values as parameters in their monitor workflow action, allowing the monitor to filter log entries based on matching OID values.

#### Release Note

OID extensions and associated values are now supported as parameters in the rekor-monitor workflow task.

#### Documentation

New OID arguments will be reflected in the rekor-monitor README in additional PRs, which will add support for OIDMatchers as CLI and GitHub Actions arguments in both named fields and in dot notation.

### DIFF
--- a/pkg/rekor/identity.go
+++ b/pkg/rekor/identity.go
@@ -85,12 +85,14 @@ type MonitoredValues struct {
 
 // IdentityEntry holds a certificate subject, issuer, and log entry metadata
 type IdentityEntry struct {
-	CertSubject string
-	Issuer      string
-	Fingerprint string
-	Subject     string
-	Index       int64
-	UUID        string
+	CertSubject    string
+	Issuer         string
+	Fingerprint    string
+	Subject        string
+	Index          int64
+	UUID           string
+	OIDExtension   asn1.ObjectIdentifier
+	ExtensionValue string
 }
 
 func (e *IdentityEntry) String() string {
@@ -170,14 +172,16 @@ func MatchedIndices(logEntries []models.LogEntry, mvs MonitoredValues) ([]Identi
 
 			for _, monitoredOID := range mvs.OIDMatchers {
 				for _, cert := range certs {
-					match, err := oidMatchesPolicy(cert, monitoredOID.ObjectIdentifier, monitoredOID.ExtensionValues)
+					match, oid, extValue, err := oidMatchesPolicy(cert, monitoredOID.ObjectIdentifier, monitoredOID.ExtensionValues)
 					if err != nil {
 						return nil, fmt.Errorf("error with policy matching for UUID %s at index %d: %w", uuid, *entry.LogIndex, err)
 					}
 					if match {
 						matchedEntries = append(matchedEntries, IdentityEntry{
-							Index: *entry.LogIndex,
-							UUID:  uuid,
+							Index:          *entry.LogIndex,
+							UUID:           uuid,
+							OIDExtension:   oid,
+							ExtensionValue: extValue,
 						})
 					}
 				}
@@ -351,22 +355,22 @@ func certMatchesPolicy(cert *x509.Certificate, expectedSub string, expectedIssue
 	return subjectMatches && issuerMatches, matchedSub, issuer, nil
 }
 
-// oidMatchesPolicy returns true if a certificate contains both a given OID field
-// and a matching value associated with that field
-func oidMatchesPolicy(cert *x509.Certificate, oid asn1.ObjectIdentifier, extensionValues []string) (bool, error) {
+// oidMatchesPolicy returns if a certificate contains both a given OID field and a matching value associated with that field
+// if true, it returns the OID extension and extension value that were matched on
+func oidMatchesPolicy(cert *x509.Certificate, oid asn1.ObjectIdentifier, extensionValues []string) (bool, asn1.ObjectIdentifier, string, error) {
 	extValue, err := getExtension(cert, oid)
 	if err != nil {
-		return false, fmt.Errorf("error getting extension value: %w", err)
+		return false, nil, "", fmt.Errorf("error getting extension value: %w", err)
 	}
 	if extValue == "" {
-		return false, nil
+		return false, nil, "", nil
 	}
 
 	for _, extensionValue := range extensionValues {
 		if extValue == extensionValue {
-			return true, nil
+			return true, oid, extValue, nil
 		}
 	}
 
-	return false, nil
+	return false, nil, "", nil
 }
--- a/pkg/rekor/identity.go
+++ b/pkg/rekor/identity.go
@@ -57,6 +57,12 @@ type CertificateIdentity struct {
 	Issuers     []string `yaml:"issuers"`
 }
 
+// OIDMatcher holds an OID field and a list of values to match on
+type OIDMatcher struct {
+	ObjectIdentifier asn1.ObjectIdentifier `yaml:"objectIdentifier"`
+	ExtensionValues  []string              `yaml:"extensionValues"`
+}
+
 // MonitoredValues holds a set of values to compare against a given entry
 type MonitoredValues struct {
 	// CertificateIdentities contains a list of subjects and issuers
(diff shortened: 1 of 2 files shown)


pkg/rekor/identity_test.go
```diff
@@ -461,6 +464,119 @@ func TestMatchedIndicesForSubjects(t *testing.T) {
...
+	pemCert, _ := cryptoutils.MarshalCertificateToPEM(leafCert)
+
+	payload := []byte{1, 2, 3, 4}
+	sig, err := signer.SignMessage(bytes.NewReader(payload))
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	hashedrekord := &hashedrekord_v001.V001Entry{}
+	h := sha256.Sum256(payload)
```
mihaimaruseac, line 499: For readability, should we use `hash` instead of `h` here and below?

pkg/rekor/identity_test.go
```diff
@@ -461,6 +464,119 @@ func TestMatchedIndicesForSubjects(t *testing.T) {
...
+		t.Fatal(err)
+	}
+	leaf, err := entry.Canonicalize(context.Background())
+	if err != nil {
+		t.Fatal(err)
+	}
+	integratedTime := time.Now()
+	logIndex := 1234
+	uuid := "123-456-123"
+	e := models.LogEntryAnon{
```
mihaimaruseac, line 520: Readability nit: let's rename this to `anonLogEntry` instead of `e`, or something like that?

pkg/rekor/identity_test.go
```diff
@@ -461,6 +464,119 @@ func TestMatchedIndicesForSubjects(t *testing.T) {
...
+		LogIndex:       swag.Int64(int64(logIndex)),
+	}
+	logEntry := models.LogEntry{uuid: e}
+
+	// match to oid with matching extension value
+	matches, err := MatchedIndices([]models.LogEntry{logEntry}, MonitoredValues{
+		OIDMatchers: []OIDMatcher{
+			{
+				ObjectIdentifier: asn1.ObjectIdentifier{1, 3, 6, 1, 4, 1, 57264, 1, 9},
+				ExtensionValues:  []string{"test cert value"},
```
mihaimaruseac, line 532: Let's reuse the values from lines 471, 472 so when those get changed we don't have to change twice

pkg/rekor/identity_test.go
```diff
@@ -461,6 +464,119 @@ func TestMatchedIndicesForSubjects(t *testing.T) {
...
+				ObjectIdentifier: asn1.ObjectIdentifier{1, 3, 6, 1, 4, 1, 57264, 1, 9},
+				ExtensionValues:  []string{"wrong"},
+			},
+		}})
+	if err != nil {
+		t.Fatalf("expected no error, got %v", err)
+	}
+	if len(matches) != 0 {
+		t.Fatalf("expected no matches, got %d", len(matches))
+	}
```
mihaimaruseac, line 561: I'd recommend moving this to a separate test (and same for the test below).

The idea is that if the above test fails (for what got inserted), this part of the test will not be tested. So, the dev will fix the previous assertion, run the test and then get another failure instead of seeing both of them at the same time.

mihaimaruseac: On the other end, I see that other test assertions are still grouped together, so maybe let's follow the same style. Let's see what @haydentherapper thinks here, I'm ok with both smaller tests and with following the existing style.

mihaimaruseac (commented):
This looks good to me, just some comments regarding the tests

mihaimaruseac (dismissed):

pkg/rekor/identity.go
```diff
@@ -158,6 +167,21 @@ func MatchedIndices(logEntries []models.LogEntry, mvs MonitoredValues) ([]Identi
...
+
+			for _, monitoredOID := range mvs.OIDMatchers {
+				for _, cert := range certs {
+					match, err := oidMatchesPolicy(cert, monitoredOID.ObjectIdentifier, monitoredOID.ExtensionValues)
+					if err != nil {
+						return nil, fmt.Errorf("error with policy matching for UUID %s at index %d: %w", uuid, *entry.LogIndex, err)
+					}
+					if match {
+						matchedEntries = append(matchedEntries, IdentityEntry{
+							Index: *entry.LogIndex,
```
haydentherapper, line 179: Can we also update `IdentityEntry` to add fields to specify what OID it was matched on and what that OID's value was? This will be needed to use `IdentityEntry` to know why there was a match.

pkg/rekor/identity_test.go
haydentherapper: I'm supportive of refactoring these tests - Mihai, did you have any good pointers for stylistic Go testing? IIRC, the preference is using tables like https://go.dev/wiki/TableDrivenTests, which I haven't done here when I wrote these tests.

With that said, let's keep any testing refactors in a dedicated PR.

mihaimaruseac: Agree with refactoring most tests in a separate PR. Here I suggested splitting the test added in the PR. But let's do all test refactoring in a dedicated PR to not keep this on for too long.

For resources, I know about https://blog.jetbrains.com/go/2022/11/22/comprehensive-guide-to-testing-in-go/

linus-sun: ah, got it- splitted the matchedIndicesFailures tests into separate tests for now, and will work on refactoring tests overall in a separate PR post merging these in if that works. thanks for the resources, will look through them once i begin the refactoring process

pkg/rekor/identity.go
```diff
@@ -313,3 +354,23 @@ func certMatchesPolicy(cert *x509.Certificate, expectedSub string, expectedIssue
...
+// oidMatchesPolicy returns if a certificate contains both a given OID field and a matching value associated with that field
+// if true, it returns the OID extension and extension value that were matched on
+func oidMatchesPolicy(cert *x509.Certificate, oid asn1.ObjectIdentifier, extensionValues []string) (bool, asn1.ObjectIdentifier, string, error) {
+	extValue, err := getExtension(cert, oid)
+	if err != nil {
+		return false, nil, "", fmt.Errorf("error getting extension value: %w", err)
+	}
+	if extValue == "" {
+		return false, nil, "", nil
+	}
```
mihaimaruseac, line 367: Should we move this check first in the function? We don't need to `getExtension` if we can return an empty/default tuple

pkg/rekor/identity.go
```diff
@@ -72,16 +78,21 @@ type MonitoredValues struct {
...
+	Issuer         string
+	Fingerprint    string
+	Subject        string
+	Index          int64
+	UUID           string
+	OIDExtension   asn1.ObjectIdentifier
+	ExtensionValue string
 }
 
 func (e *IdentityEntry) String() string {
```
mihaimaruseac, line 98: Note: we need to extend this to include the new fields (see the range in line 89). I think it's ok to do this in a new PR

pkg/rekor/identity.go
linus-sun: i think we're checking if the extension field returned by getExtension is empty and returning on the empty case here- i can add checks for if the input args of oid/extensionValues are empty if that would be preferable?

pkg/rekor/identity.go
linus-sun: creating a separate PR for this!

pkg/rekor/identity.go
mihaimaruseac: Sorry, I misread, somehow I confused `extValue` with `extensionValues`

mihaimaruseac (approved):
This LGTM to me, thank you!

haydentherapper (approved):
Great work!

PR #458, opened 2024-09-18
Linussun/parse json

linus-sun:


pkg/rekor/identity.go
```diff
@@ -87,14 +87,26 @@ type MonitoredValues struct {
 
 // IdentityEntry holds a certificate subject, issuer, OID extension and associated value, and log entry metadata
 type IdentityEntry struct {
```
linus-sun, line 89: I'm also open to a refactoring of IdentityEntry into `pkg/identity`, which would move this code into there as well- more broadly, what should the specifications be determining if code should go in `pkg/rekor/identity` vs `pkg/identity`?

mihaimaruseac (approved):
Thank you!

haydentherapper (commented):
Looks great!

haydentherapper (approved):

PR #459, opened 2024-09-19
implement API method to create GitHub issues notifications on found identity entries

linus-sun:
#### Summary

As specified [here](https://docs.google.com/document/d/1A5OdqgJaiWcGvnMvqIfDHu94bCCYAseQEav7YXDwiBs/edit?resourcekey=0-GS2fXhOS7nMFWvnLfkmDKA&tab=t.0#heading=h.xgjl2srtytjt), this PR implements the API method `GitHubIssueInput.Send` to create a GitHub issue as a notification method for a rekor-monitor run, which takes in a list of found identities and their associated log entries and outputs it within the created GitHub issue body as follows:

image

#### Release Note

New API method to create a GitHub issue in the specified repository with a given assignee username, which details a list of monitored identities with matching log entries

#### Documentation

Documentation will be updated accordingly to surface this new method once it is implemented within CLI


mihaimaruseac:
I think this builds on top of #458? Let's wait for that one.

Meanwhile, should we try to get coverage at higher values? I'm concerned that project coverage is at ~50%, though also it seems the codecov plugin has ran at commits that are too old :-?

pkg/notifications/notifications.go
```diff
@@ -0,0 +1,91 @@
...
+
+// NotificationPlatform provides the Send() method to handle alerting logic
+// for the respective notification platform extending the interface.
+type NotificationPlatform interface {
+	Send([]identity.MonitoredIdentity) error
+}
+
+// GitHubIssueInput extends the NotificationPlatform interface to support found identity
+// notification via creating new GitHub issues in a given repo.
+type GitHubIssueInput struct {
```
haydentherapper, line 47: Could we move the implementations of the interface to their own files? Let's keep this file as just the definition of the interface plus anything else common between all implementations

pkg/notifications/notifications.go
```diff
@@ -0,0 +1,91 @@
...
+	GitHubAssigneeUsername string
```
haydentherapper, line 48: Small naming improvement, since these are all grouped under a struct GitHub*, you could drop the prefix 'GitHub' for each field.

pkg/notifications/notifications.go
```diff
@@ -0,0 +1,91 @@
...
+	if err != nil {
+		return err
+	}
+	var client *github.Client
+	if gitHubIssueInput.GitHubClient == nil {
+		client = github.NewClient(nil).WithAuthToken(gitHubIssueInput.AuthenticationToken)
+	} else {
+		client = gitHubIssueInput.GitHubClient
+	}
+	ctx := context.Background()
```
haydentherapper, line 80: We should add `context.Context` as input for the `Send` interface so callers can choose if they want to define their own context.

linus-sun:
> I think this builds on top of #458? Let's wait for that one.
> 
> Meanwhile, should we try to get coverage at higher values? I'm concerned that project coverage is at ~50%, though also it seems the codecov plugin has ran at commits that are too old :-?

https://github.com/sigstore/rekor-monitor/pull/462 should address some of the lowered coverage (a lot of it was due to `pkg/extensions`, specifically the OID extensions; a test that checks all OID fields improved coverage significantly, and I'll continue to address coverage as work continues within rekor-monitor).

mihaimaruseac (approved):

pkg/notifications/github_issues.go
```diff
@@ -0,0 +1,84 @@
...
+	"strings"
+	"time"
+
+	"github.com/google/go-github/v65/github"
+	"github.com/sigstore/rekor-monitor/pkg/identity"
+)
+
+var (
+	notificationPlatformGitHubIssueBodyHeaderText = "Rekor-monitor found the following pairs of monitored identities and matching log entries: "
+	notificationPlatformGitHubIssueLabels         = []string{"rekor-monitor", "automatically generated"}
```
haydentherapper, line 36: What happens if the labels don't exist in the repo? Do we need to document that we assume these labels exist, or are the labels created automatically?

pkg/notifications/github_issues.go
```diff
@@ -0,0 +1,84 @@
...
+)
+
+// GitHubIssueInput extends the NotificationPlatform interface to support found identity
+// notification via creating new GitHub issues in a given repo.
+type GitHubIssueInput struct {
+	AssigneeUsername string
+	OwnerUsername    string
```
haydentherapper, line 43: Is this the org or who is filing the issue? If it's the org, I'd change to `OrgName`

haydentherapper (commented):
Great work on this, just a few small comments!

pkg/notifications/github_issues.go
linus-sun: this is user who's filing the issue, gonna change to `RepositoryOwner` to make more clear

pkg/notifications/github_issues.go
linus-sun: from testing, it seems that the labels are created automatically

pkg/notifications/github_issues.go
haydentherapper: Ah, I realize what threw me off - looking at https://docs.github.com/en/rest/issues/issues?apiVersion=2022-11-28#create-an-issue, the "owner" a repository is either a user or an org. In the case of an org, "username" wouldn't be correct since the org isn't a user. WDYT about just `RepositoryOwner`?

linus-sun: sounds good to me!

haydentherapper (approved):

PR #460, opened 2024-09-20
refactor extensions and identity packages

linus-sun:
#### Summary

This PR refactors the `extensions`, `pkg/identity`, and `pkg/rekor` packages to move structs around OID extensions and rekor identities into more suitable respective packages, in the goal that `pkg/rekor/identity` doesn't suffer from bloating, and building more features out of rekor-monitor is easier in terms of splitting work into separate packages (ex. future work on expanding found identity notification and notification platforms will be easier to place in its own `notification` package).

#### Release Note

NONE

#### Documentation

N/A

### DIFF
--- a/pkg/identity/identity.go
+++ b/pkg/identity/identity.go
@@ -16,40 +16,63 @@ package identity
 
 import (
 	"encoding/asn1"
-	"errors"
 	"strconv"
 	"strings"
+
+	"github.com/sigstore/rekor-monitor/pkg/fulcio/extensions"
 )
 
-// OIDMatcher holds an OID field and a list of values to match on
-type OIDMatcher struct {
-	ObjectIdentifier asn1.ObjectIdentifier `yaml:"objectIdentifier"`
-	ExtensionValues  []string              `yaml:"extensionValues"`
+// CertificateIdentity holds a certificate subject and an optional list of identity issuers
+type CertificateIdentity struct {
+	CertSubject string   `yaml:"certSubject"`
+	Issuers     []string `yaml:"issuers"`
 }
 
-// CustomOID holds an OID field represented in dot notation and a list of values to match on
-type CustomExtension struct {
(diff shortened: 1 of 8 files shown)


pkg/identity/identity.go
```diff
@@ -16,40 +16,63 @@ package identity
...
-	nodes := strings.Split(oid, ".")
-	objectIdentifier := make([]int, len(nodes))
-	for i, node := range nodes {
-		if strings.TrimSpace(node) == "" {
-			return nil, errors.New("could not parse object identifier: no characters between two dots")
-		}
-		intNode, err := strconv.Atoi(node)
-		if err != nil {
-			return nil, err
+// IdentityEntry holds a certificate subject, issuer, OID extension and associated value, and log entry metadata
```
linus-sun, line 57: linter prompts that this import becomes `identity.IdentityEntry`, which is redundant and proposes renaming it `Entry`; I think that removes clarity but am inclined to agree about the redundancy. I could potentially see something like `LogIdentityEntry`- is the redundancy an issue we want to deal with, and if so, does something like this work? open to any more suitable renames as well

mihaimaruseac: I tend to agree with the redundancy. What about `CertificateEntry`?

mihaimaruseac (approved):

pkg/identity/identity.go
linus-sun: i believe technically an entry here does not need to contain any information tied to a certificate (eg. could include an SSH/PGP key related email address under the Subjects field), so that could result in reduced clarity- If something like `RekorLogEntry` or `RekorEntry` seems reasonable, could rename it as such

haydentherapper (approved):
Looks great!

As a small thing, can you update the PR description with what changes you made outside of moving code around, e.g renamings, just so it's easier to know at a glance what other refactors have been done?

PR #471, opened 2024-10-02
Implement separate methods for log consistency check + identity search

linus-sun:
#### Summary

This PR implements split methods to handle separate logic for performing a log consistency check and identity search within rekor-monitor, following work outlined [in this document](https://docs.google.com/document/d/15XEwkUDy2Wp9oSpazD18vMMbeo-UDpqpzPAiiMkGvFk/edit?resourcekey=0-l0iCV7L2HCa8xC3XCEz8rg&tab=t.0). This PR duplicates code, maintaining the current workflow and implementing these separate components as unused methods; a future PR will migrate the CLI and GitHub Actions reusable workflow to these newly implemented methods.

#### Release Note

New endpoints have been added to handle log consistency checks and identity searches separately

#### Documentation

Documentation will be updated once rekor-monitor is migrated to using these independent methods.


pkg/rekor/identity.go
```diff
@@ -335,3 +353,65 @@ func oidMatchesPolicy(cert *x509.Certificate, oid asn1.ObjectIdentifier, extensi
 
 	return false, nil, "", nil
 }
+
+func LogIdentitySearch(config LogIdentitySearchConfiguration, rekorClient *gclient.Rekor) error {
+	if (config.StartIndex == nil && config.EndIndex != nil) || (config.StartIndex != nil && config.EndIndex == nil) {
```
mihaimaruseac, line 358: Can we use previous checkpoint if only start is not given? Can we use current checkpoint if end is not given?

If we can, it would simplify this, as we would have a single method to convert from start index pointer to the value given by user or the last checkpoint and similarly for the end index (with current checkpoint)

haydentherapper (changes requested):
For a refactor, I'd prefer if we can break this into smaller chunks. This has a few benefits:

* It's obvious where the refactor is happening vs a new control flow being added
* We can add tests as we go - with this refactor, there are no tests (since there were no tests in the main file, this doesn't change coverage, but we should try to fix this as we refactor)
* We can reduce duplication - I know you mentioned intentionally having these functions duplicated, which I think is fine for a refactor, but between these two added functions there is duplication that we should refactor out.
* We avoid large green and large red PRs, which is where bugs slip in or functionality is accidentally dropped

I'm a fan of https://refactoring.guru/refactoring/catalog which discusses "code smells", aka issues that we want to resolve in the code - in this case, long functions and coupled logic - and some suggested refactoring techniques. A lot of these suggestions are focused on OOP which doesn't always apply in Go, but I still find this guide generally helpful.

What do you think about splitting this up into a few PRs where we start to pull out logic from the large function in main into dedicated, tested functions in verification.go?

linus-sun:
> What do you think about splitting this up into a few PRs where we start to pull out logic from the large function in main into dedicated, tested functions in verification.go?

I think this sounds good- I'll work on splitting up the primary workflow into smaller components, first extracting out the identity search and then decomposing the consistency check.

PR #491, opened 2024-10-16
add identity search workflow

linus-sun:
#### Summary

As part of work [outlined here](https://docs.google.com/document/d/15XEwkUDy2Wp9oSpazD18vMMbeo-UDpqpzPAiiMkGvFk/edit?resourcekey=0-l0iCV7L2HCa8xC3XCEz8rg&tab=t.0#heading=h.xgjl2srtytjt), this PR adds a separate identity monitoring workflow and corresponding end-to-end test within Rekor-monitor. This workflow takes in a list of associated identities to monitor for, and searches through a range of logs to write any found identities to disk. A future PR will refactor the identity search component out of the current consistency check and reusable monitoring workflow and update documentation to demonstrate how to use the separate consistency check and identity search workflows.

#### Release Note

NONE

#### Documentation

N/A - a future PR that refactors this to be runnable as a separate workflow will update documentation accordingly.

### DIFF
--- a/cmd/monitor/config.go
+++ b/cmd/monitor/config.go
@@ -1,0 +1,37 @@
+//
+// Copyright 2024 The Sigstore Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"time"
+
+	"github.com/sigstore/rekor-monitor/pkg/identity"
+	"github.com/sigstore/rekor-monitor/pkg/notifications"
+)
+
+type IdentityMonitorConfiguration struct {
+	StartIndex                *int                                    `yaml:"startIndex"`
+	EndIndex                  *int                                    `yaml:"endIndex"`
+	MonitoredValues           identity.MonitoredValues                `yaml:"monitoredValues"`
+	ServerURL                 string                                  `yaml:"serverURL"`
+	OutputIdentitiesFile      string                                  `yaml:"outputIdentities"`
+	LogInfoFile               string                                  `yaml:"logInfoFile"`
+	GitHubIssue               notifications.GitHubIssueInput          `yaml:"githubIssue"`
+	EmailNotificationSMTP     notifications.EmailNotificationInput    `yaml:"emailNotificationSMTP"`
+	EmailNotificationMailgun  notifications.MailgunNotificationInput  `yaml:"emailNotificationMailgun"`
+	EmailNotificationSendGrid notifications.SendGridNotificationInput `yaml:"emailNotificationSendGrid"`
+	Interval                  *time.Duration                          `yaml:"interval"`
+}
--- a/cmd/monitor/config.go
+++ b/cmd/monitor/config.go
@@ -29,6 +29,7 @@ type IdentityMonitorConfiguration struct {
 	ServerURL                 string                                  `yaml:"serverURL"`
 	OutputIdentitiesFile      string                                  `yaml:"outputIdentities"`
 	LogInfoFile               string                                  `yaml:"logInfoFile"`
+	IdentityMetadataFile      *string                                 `yaml:"identityMetadataFile"`
 	GitHubIssue               notifications.GitHubIssueInput          `yaml:"githubIssue"`
 	EmailNotificationSMTP     notifications.EmailNotificationInput    `yaml:"emailNotificationSMTP"`
 	EmailNotificationMailgun  notifications.MailgunNotificationInput  `yaml:"emailNotificationMailgun"`
--- a/cmd/monitor/main.go
+++ b/cmd/monitor/main.go
@@ -1,0 +1,143 @@
+//
+// Copyright 2021 The Sigstore Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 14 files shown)


mihaimaruseac (approved):
Nice!

cmd/monitor/config.go
```diff
@@ -0,0 +1,34 @@
...
+// limitations under the License.
+
+package main
+
+import (
+	"github.com/sigstore/rekor-monitor/pkg/identity"
+	"github.com/sigstore/rekor-monitor/pkg/notifications"
+)
+
+type IdentityMonitorConfiguration struct {
```
haydentherapper, line 23: Should we also include `once` so this either runs indefinitely or only for one iteration? 

If so, this would need to be restructured a bit, so that we continually pull in the start index from the last checkpoint that was recorded.

haydentherapper: related, I think if you specify start/end, that should effectively be the same as `once`, as a one-off search across indices (which could be used if you are running the search for the first time, for example, or catching up)

linus-sun: I was considering how to include once but I think that makes sense for constantly pulling in the last checkpoint, will do

haydentherapper: We should also decide between the indefinite loop being in the cmd package or in the inner API. Right now, the functionality is embedded in the API, and I wouldn’t expect API users to need to loop indefinitely. Also it’s not possible to test an endless loop, so we can maintain good coverage in the API where the cmd package isn’t expected to have coverage.

linus-sun: looping in the cmd package definitely makes more sense- will implement that way in this PR, and refactor consistency check as such in #492

cmd/monitor/main.go
```diff
@@ -0,0 +1,143 @@
...
+	if err := yaml.Unmarshal([]byte(configString), &config); err != nil {
+		log.Fatalf("error parsing identities: %v", err)
+	}
+
+	rekorClient, err := client.GetRekorClient(config.ServerURL, client.WithUserAgent(strings.TrimSpace(fmt.Sprintf("rekor-monitor/%s (%s; %s)", version.GetVersionInfo().GitVersion, runtime.GOOS, runtime.GOARCH))))
+	if err != nil {
+		log.Fatalf("getting Rekor client: %v", err)
+	}
+
+	if config.StartIndex == nil || config.EndIndex == nil {
```
haydentherapper, line 72: This loop is missing writing the end index or the latest checkpoint to a file, so for a future invocation of the binary, we'd start from the same index, right?

Given that we don't actually need checkpoints, I wonder if we can simplify this if we instead propose a very basic structure to hold the last successfully verified index like `{last_index: x}` (and in the future, it could also include all indices that could not be verified, the ones we will skip over).

Then it becomes something like:

* Run all of the below in a loop:
  * If you don't specify a start index, read from file
  * If you don't specify an end index, ask for the latest checkpoint
  * If end < start, fail early
  * Run `IdentitySearch`
  * Persist end index in file
  * If `once` or you specified an end index, exit (WDYT about the latter part? I think if you say you want to check only up to X, then that is implicitly asking to be run once)

linus-sun: Agree with all of this- I think `once` can default to true, and in the case that a user does want to initially check up to the end index but continue to leave the monitor running, that option can be left open to them

pkg/rekor/identity.go
```diff
@@ -351,25 +371,55 @@ func writeIdentitiesBetweenCheckpoints(logInfo *models.LogInfo, prevCheckpoint *
...
+
+			if err := file.WriteIdentity(outputIdentitiesFile, idEntry); err != nil {
+				return fmt.Errorf("failed to write entry: %v", err)
+			}
 		}
-		idEntries, err := MatchedIndices(entries, monitoredValues)
+	}
+
+	// TODO: idMetadataFile currently takes in a string pointer to not cause a regression in the current reusable monitoring workflow.
+	// Once the reusable monitoring workflow is split into a consistency check and identity search, idMetadataFile should always take in a string value.
```
linus-sun, line 399: Potentially could also leave as an optional in case users don't want the extra file write/use case of identity metadata file

cmd/monitor/main.go
```diff
@@ -0,0 +1,135 @@
...
+			if config.EndIndex == nil {
+				config.EndIndex = &checkpointEndIndex
+			}
+		}
+
+		if *config.StartIndex >= *config.EndIndex {
+			fmt.Fprintf(os.Stderr, "start index %d must be strictly less than end index %d", *config.StartIndex, *config.EndIndex)
+		}
+
+		// TODO: This should subsequently read from the identity metadata file to fetch the latest index.
```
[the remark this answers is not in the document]
haydentherapper: I'm good with addressing this in a follow up. However there's a few issues we need to address in this PR before we can merge.

There are effectively three sources for a `startIndex` value - 1) `config.StartIndex`, 2) the previous checkpoint size (in the future, to be replaced by the identity metadata), 3) the end index from the last run. You're missing the last of these. If you don't handle this, then if `config.Start/EndIndex` is set, you'll always use the same value.

For an `endIndex` value, there are two sources - 1) `config.EndIndex`, or the latest checkpoint size. You handle both, but as I mentioned previously, I think if you set `config.EndIndex`, we should only execute the loop once, as the client is stating an end. So if `config.EndIndex` is set, either set `once` to true or update the condition to end early to `*once || config.EndIndex != nil`.

 What do you think about these suggestions? so tl;dr:

1. Persist the end index between each run and set that as the start index for the next run
2. Exit early if `config.EndIndex != nil`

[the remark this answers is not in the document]
linus-sun: https://github.com/sigstore/rekor-monitor/pull/491/files#diff-fd15e03959354998b048bc5aec461761a2d24e159194ad4bf6d89dd9f7e3b58cR131-R136 this should handle exiting early if config.EndIndex has been set as well as persisting the endIndex as the startIndex for the next run

cmd/monitor/main.go
```diff
@@ -0,0 +1,138 @@
...
+		if err != nil {
+			fmt.Fprintf(os.Stderr, "failed to successfully complete identity search: %v", err)
+			return
+		}
+
+		if *once || config.EndIndex != nil {
+			return
+		}
+
+		config.StartIndex = config.EndIndex
```
haydentherapper, line 135: You'll need a variable initialized to `config.Start/EndIndex` at the beginning of the loop. Otherwise, `config.EndIndex` will get set in the conditional and then this will exit early. Does that seem correct?

linus-sun: thanks for catching that, should be updated now

haydentherapper (approved):

PR #502, opened 2024-10-23
verify ct consistency code

linus-sun:


pkg/rekor/verifier.go
```diff
@@ -23,6 +23,8 @@ import (
 	"os"
 	"time"
 
+	ct "github.com/google/certificate-transparency-go"
```
haydentherapper, line 26: Can we create another package, like `pkg/ct`, for handling CT verification, since this isn't for rekor?

pkg/util/file/file.go
```diff
@@ -52,6 +54,55 @@ func ReadLatestCheckpoint(logInfoFile string) (*util.SignedCheckpoint, error) {
...
+
+	if err := scanner.Err(); err != nil {
+		return nil, err
+	}
+
+	return &checkpoint, nil
+}
+
+// WriteCTSignedTreeHead writes a signed tree head to a given log file
+func WriteCTSignedTreeHead(sth *ct.SignedTreeHead, logInfoFile string) error {
```
haydentherapper, line 87: Let's take this opportunity to stop writing all checkpoints, and just overwrite the file with the latest checkpoint. This will simplify reading as well.

pkg/rekor/verifier.go
```diff
@@ -100,6 +105,23 @@ func VerifyConsistencyCheckInputs(interval *time.Duration, logInfoFile *string,
 	return nil
 }
 
+func verifyCertificateTransparencyConsistency(logInfoFile string, logClient *ctclient.LogClient, signedTreeHead *ct.SignedTreeHead) error {
```
haydentherapper, line 108: We'll also need signature verification for the previous STH and current STH, like with https://github.com/sigstore/rekor-monitor/blob/d7458ab1ec52524bf11057d4c2e38bf07b6a5a5c/pkg/rekor/verifier.go#L56. Checkpoints are signed, which provides integrity for the checkpoint.

We can hardcode the public key for now from https://github.com/sigstore/root-signing/blob/main/targets/ctfe_2022.pub, but we should add a TODO to fetch it from our TUF repository like what we need to do for rekor.

haydentherapper (changes requested):

pkg/ct/verifier.go
```diff
@@ -0,0 +1,75 @@
...
+
+		if err != nil {
+			return fmt.Errorf("error loading public key: %v", err)
+		}
+		logClient.Verifier = &ct.SignatureVerifier{
+			PubKey: pubKey,
+		}
+	}
+
+	err = logClient.Verifier.VerifySTHSignature(*prevSTH)
```
haydentherapper, line 54: nit: For consistency with below, do you need `logClient.Verifier` vs just `logClient`?

pkg/ct/verifier_test.go
```diff
@@ -0,0 +1,114 @@
...
+	return serveHandlerAt(t, path, func(w http.ResponseWriter, _ *http.Request) {
+		if _, err := fmt.Fprint(w, rsp); err != nil {
+			t.Fatal(err)
+		}
+	})
+}
+
+// Test VerifyCertificateTransparencyConsistency
+func TestVerifyCertificateTransparencyConsistency(t *testing.T) {
+	// TODO: placeholder test, fill this out with mock CT Log client
```
haydentherapper, line 66: just to check, is there anything else you wanted to test in this PR, or adding in a later one?

haydentherapper (commented):
This looks great!

pkg/ct/verifier_test.go
linus-sun: Adding in a later one once security-transparency responds w.r.t. setting up end-to-end testing- validating this function requires ECDSA signing the STH using the ctfe pub/priv key- is it possible to mock that/generate one to pass into the function here, like as done [here](https://github.com/google/certificate-transparency-go/blob/517bc29902ecdd3e0ae01bcd1b588cfb31f11bed/client/logclient_test.go#L52-L55)?

haydentherapper: When I set up testing awhile ago, I hardcoded a response - https://github.com/sigstore/fulcio/blob/25bdda2ef4094ea73ae864f3c89f924570e637fd/pkg/server/grpc_server_test.go#L2291

e2e tests are probably the most straightforward way to test without copying a huge amount of code.

haydentherapper (approved):

haydentherapper (approved):

PR #512, opened 2024-10-29
Add support for triggering notifications off of identity monitor

linus-sun:
#### Summary

This PR adds support for triggering notifications off of the identity monitor workflow. Currently, GitHub issue notifications and email notifications (via SMTP, Mailgun, and SendGrid) are supported; future notification handlers can be implemented via the identity monitor configuration method implemented in this PR. 

#### Release Note

NONE

#### Documentation

N/A


pkg/rekor/identity.go
```diff
@@ -403,11 +403,13 @@ func IdentitySearch(startIndex int, endIndex int, rekorClient *client.Rekor, mon
 		}
 		err = file.WriteIdentityMetadata(*idMetadataFile, idMetadata)
 		if err != nil {
-			return fmt.Errorf("failed to write id metadata: %v", err)
+			return nil, fmt.Errorf("failed to write id metadata: %v", err)
 		}
 	}
 
-	return nil
+	identities := identity.CreateIdentitiesList(monitoredValues)
```
haydentherapper, line 410: What is the reason for this change? How will the identity strings be used?

cmd/monitor/config.go
```diff
@@ -16,23 +16,58 @@
...
+	OutputIdentitiesFile      string                                   `yaml:"outputIdentities"`
+	LogInfoFile               string                                   `yaml:"logInfoFile"`
+	IdentityMetadataFile      *string                                  `yaml:"identityMetadataFile"`
+	GitHubIssue               *notifications.GitHubIssueInput          `yaml:"githubIssue"`
+	EmailNotificationSMTP     *notifications.EmailNotificationInput    `yaml:"emailNotificationSMTP"`
+	EmailNotificationMailgun  *notifications.MailgunNotificationInput  `yaml:"emailNotificationMailgun"`
+	EmailNotificationSendGrid *notifications.SendGridNotificationInput `yaml:"emailNotificationSendGrid"`
+	Interval                  *time.Duration                           `yaml:"interval"`
+}
+
```
haydentherapper, line 41: What do you think about having a "pool" of notification providers? This would be splitting `TriggerNotifications` into two functions:

* Something like `CreateNotificationPlatformPool` which outputs `[]NotificationPlatform`, which you would populate based on what's set in the configuration
* `TriggerNotifications` which would iterate through the list and call `Send` on each

Functionally, these are equivalent, but it's a bit more idiomatic to use the interface.

pkg/rekor/identity.go
linus-sun: the output of CreateIdentitiesList is used in [CreateMonitoredIdentities](https://github.com/sigstore/rekor-monitor/blob/b3b5cb9ee132ff15f784b1b4c00413766a3faaca/pkg/identity/identity.go#L99) to help create the structured output of found identity entries, which is passed in as the input for any attached notification platforms to create notifications off of

cmd/monitor/config.go
```diff
@@ -16,23 +16,58 @@
...
+func CreateNotificationPool(config IdentityMonitorConfiguration) []notifications.NotificationPlatform {
```
haydentherapper, line 42: Love this construct!

Can we move these functions to the notifications package, and add tests? You can create a fake NotificationPlatform and verify that `Send` is called after `CreateNotificationPool` and `TriggerNotifications` is called.

pkg/rekor/identity.go
haydentherapper: Ah thanks, forgot we are grouping found entries by id. Looks good!

pkg/identity/identity.go
```diff
@@ -94,6 +94,26 @@ func PrintMonitoredIdentities(monitoredIdentities []MonitoredIdentity) ([]byte,
 	return jsonBody, nil
 }
 
+// CreateIdentitiesList takes in a MonitoredValues input and returns a list of all currently monitored identities.
+// It returns a list of strings.
+func CreateIdentitiesList(mvs MonitoredValues) []string {
```
haydentherapper, line 99: Test for func?

haydentherapper (approved):
looks great!

PR #513, opened 2024-10-29
Refactor consistency check workflow

linus-sun:
#### Summary

This PR does the following:
- Renames the `verifier` task to `rekor_consistency` to more accurately reflect its functionality
- refactors `RunConsistencyCheck` to take the looping functionality out of the check itself and moves it into the workflow
- refactors `RunConsistencyCheck` to move its identity monitor functionality into a separate function, `writeIdentitiesBetweenCheckpoints`

This PR has been tested remotely via a GitHub reusable workflow: https://github.com/linus-sun/rekor-log-monitor/actions/runs/11582060548/job/32244210000

#### Release Note

- Workflow configuration inputs changed to take in either a yaml as a string or local file path

#### Documentation

Readme has been updated to reflect new configuration input


cmd/consistency/main.go
```diff
@@ -0,0 +1,132 @@
...
+	ticker := time.NewTicker(*config.Interval)
+	defer ticker.Stop()
+
+	// Loop will:
+	// 1. Fetch latest checkpoint and verify
+	// 2. If old checkpoint is present, verify consistency proof
+	// 3. Write latest checkpoint to file
+
+	// To get an immediate first tick
+	for ; ; <-ticker.C {
```
haydentherapper, line 121: Did you want to include a loop here? `RunConsistencyCheck` includes a loop.

I would suggest renaming the file with `git mv` and then making changes, which I believe will properly show the diff in github.

cmd/consistency/main.go
haydentherapper, line 1: Should we name this specific to rekor consistency, since we'll also have CT?

cmd/consistency/main.go
linus-sun: Removing loop from within `RunConsistencyCheck` tried renaming via `git mv`, but diff doesn't still seem to be tracked- might be because I already renamed without using `git mv`

cmd/rekor_consistency/config.go
```diff
@@ -0,0 +1,29 @@
...
+
+package main
+
+import (
+	"time"
+
+	"github.com/sigstore/rekor-monitor/pkg/identity"
+)
+
+type ConsistencyCheckConfiguration struct {
```
haydentherapper, line 24: What do you think about omitting this and sticking to CLI flags? I think that yaml is great for complex input or for server configuration, but since we expect this to be run as a tool, these can be flags.

Given we also have other flags already, I don't think we need a configuration too.

If we remove this, then we can have a `monitored-values` flag as a string and a `monitored-value-file` flag as a path to a file.

cmd/rekor_consistency/main.go
```diff
@@ -0,0 +1,132 @@
...
+
+// This main function performs a periodic root hash consistency check.
+// Upon starting, any existing latest snapshot data is loaded and the function runs
+// indefinitely to perform consistency check for every time interval that was specified.
+func main() {
+	// Command-line flags that are parameters to the verifier job
+	configFilePath := flag.String("config-file", "", "Name of the file containing the consistency check workflow configuration settings")
+	configString := flag.String("config-string", "", "Consistency check workflow configuration settings input as a string")
+	once := flag.Bool("once", false, "Perform consistency check once and exit")
+	logInfoFile := flag.String("file", "", "path to log info file")
```
haydentherapper, line 50: You can specify default values instead of an empty string, and if `--file` is set, it'll override the value.

.github/workflows/reusable_monitoring.yml
```diff
@@ -76,7 +76,7 @@ jobs:
         run: cat ${{ env.LOG_FILE }}
         # Skip on first run
         continue-on-error: true
-      - run: go run ./cmd/verifier --file ${{ env.LOG_FILE }} --once --monitored-values "${{ inputs.identities }}" --user-agent "${{ format('{0}/{1}/{2}', needs.detect-workflow.outputs.repository, needs.detect-workflow.outputs.ref, github.run_id) }}"
+      - run: go run ./cmd/consistency --file ${{ env.LOG_FILE }} --once --config-string "${{ inputs.config }}" --user-agent "${{ format('{0}/{1}/{2}', needs.detect-workflow.outputs.repository, needs.detect-workflow.outputs.ref, github.run_id) }}"
```
haydentherapper, line 79: `rekor_consistency`?

haydentherapper (commented):
Just two small comments, looks good otherwise!

cmd/rekor_consistency/config.go
haydentherapper: Do we want to address the last part of this:

> If we remove this, then we can have a monitored-values flag as a string and a monitored-value-file flag as a path to a file.

Adding a flag for passing by file as well?

go.mod
```diff
@@ -14,6 +14,7 @@ require (
 	github.com/sigstore/sigstore v1.8.10
 	github.com/wneessen/go-mail v0.5.1
 	golang.org/x/mod v0.21.0
+	gopkg.in/yaml.v2 v2.4.0
```
haydentherapper, line 17: I think this can be reverted?

cmd/rekor_consistency/main.go
```diff
@@ -55,6 +56,16 @@ func main() {
 	userAgentString := flag.String("user-agent", "", "details to include in the user agent string")
 	flag.Parse()
 
+	rekorClient, err := client.GetRekorClient(*serverURL, client.WithUserAgent(strings.TrimSpace(fmt.Sprintf("rekor-monitor/%s (%s; %s) %s", version.GetVersionInfo().GitVersion, runtime.GOOS, runtime.GOARCH, *userAgentString))))
```
haydentherapper, line 59: tiny nit, does this need to be moved? if not, it just keeps the diff smaller

haydentherapper (dismissed):

haydentherapper:
LGTM, can you rebase and run `go mod tidy` to resolve the conflict? Can you also update the PR description to reflect what's in this PR?

PR #514, opened 2024-10-30
Add functionality for finding subjects in certificate transparency logs

linus-sun:
#### Summary

This PR, [per work outlined here](https://docs.google.com/document/d/1fdO5jMTJOtMhIWs_lzWlb25olIWMvYix8LW7ugeMH8k/edit?tab=t.0#heading=h.xgjl2srtytjt), begins implementation of supporting identity monitoring within certificate transparency logs by adding support for monitoring subjects within a certificate in a certificate transparency log's entry. Functionality is tested via unit testing.

#### Release Note

NONE

#### Documentation

N/A


pkg/ct/monitor_test.go
```diff
@@ -0,0 +1,117 @@
...
+	"github.com/google/certificate-transparency-go/x509"
+	"github.com/google/certificate-transparency-go/x509/pkix"
+	"github.com/sigstore/rekor-monitor/pkg/identity"
+)
+
+const (
+	subjectName = "test-subject"
+)
+
+func TestGetCTLogEntries(t *testing.T) {
```
linus-sun, line 36: unsure this is necessary- this is effectively a copy of the same test found here https://github.com/google/certificate-transparency-go/blob/6b2be7508011bf6c81c13a9ec532754e77c9bc79/client/logclient_test.go#L157, and GetCTLogEntries is just a wrapper on this method.

pkg/identity/identity.go
```diff
@@ -55,8 +55,8 @@ type MonitoredValues struct {
 	CustomExtensions []extensions.CustomExtension `yaml:"customExtensions"`
 }
 
-// RekorLogEntry holds a certificate subject, issuer, OID extension and associated value, and log entry metadata
-type RekorLogEntry struct {
+// LogEntry holds a certificate subject, issuer, OID extension and associated value, and log entry metadata
```
linus-sun, line 58: Renamed RekorLogEntry to LogEntry to be used in both Rekor and CT log entries found via identity monitoring- would it be better to keep them as distinct structs?

pkg/ct/monitor_test.go
haydentherapper: Given it's just a wrapper, I wouldn't say the test is necessary.

pkg/identity/identity.go
haydentherapper: Let's reuse, it minimizes duplicated code.

pkg/ct/monitor.go
```diff
@@ -0,0 +1,53 @@
...
+
+func GetCTLogEntries(logClient *ctclient.LogClient, startIndex int, endIndex int) ([]ct.LogEntry, error) {
+	entries, err := logClient.GetEntries(context.Background(), int64(startIndex), int64(endIndex))
+	if err != nil {
+		return nil, fmt.Errorf("error retrieving certificate transparency log entries: %v", err)
+	}
+	return entries, nil
+}
+
+func ScanEntrySubject(logEntry ct.LogEntry, monitoredSubjects []string) ([]*identity.LogEntry, error) {
```
haydentherapper, line 35: Can we reuse [MatchedIndices](https://github.com/sigstore/rekor-monitor/blob/main/pkg/rekor/identity.go#L60) since it supports matching by many other values beyond subject?

We'll want two versions of the function, one which takes `ct.LogEntry`s and one for `models.LogEntry`.

haydentherapper (approved):

PR #527, opened 2024-11-07
Add functionality for scanning OID matchers in a CT log entry

linus-sun:
#### Summary

Per changes outlined in [this doc](https://docs.google.com/document/d/1fdO5jMTJOtMhIWs_lzWlb25olIWMvYix8LW7ugeMH8k/edit?tab=t.0#heading=h.xgjl2srtytjt), this PR implements support for scanning asn1.ObjectIdentifier extensions and matching values within a certificate transparency log. A certificate's extensions can be matched on and filtered for specific extension values, (future PRs will handle saving this output and writing to any of various notification platforms).

This PR also refactors functions related to scanning OID extensions and matching on them for specific values.

#### Release Note

NONE

#### Documentation

N/A


mihaimaruseac (dismissed):

pkg/identity/identity.go
haydentherapper, line 1: Do we need to these functions to this package? Given they are only used by the Rekor package and represent identity in the Rekor log, I'd suggest leaving them as-is.

haydentherapper: I mentioned this offline, but I think it may be possible to use [generics](https://go.dev/doc/tutorial/generics) here, defining a generic `getExtension` and other functions with a type constraint of `[T x509.Certificate | google_x509.Certificate`. If so, then it would make sense to move these to this package.

If generics aren't a fit, then I'd suggest reverting this move and keeping the Rekor identity functions in the rekor package.

pkg/ct/monitor.go
```diff
@@ -32,22 +35,44 @@ func GetCTLogEntries(logClient *ctclient.LogClient, startIndex int, endIndex int
 	return entries, nil
 }
 
-func ScanEntrySubject(logEntry ct.LogEntry, monitoredSubjects []string) ([]*identity.LogEntry, error) {
+func ScanEntryCertSubject(logEntry ct.LogEntry, monitoredSubjects []string) ([]*identity.LogEntry, error) {
 	subject := logEntry.X509Cert.Subject.String()
-	foundEntries := []*identity.LogEntry{}
+	matchedEntries := []*identity.LogEntry{}
 	for _, monitoredSub := range monitoredSubjects {
 		regex, err := regexp.Compile(monitoredSub)
```
haydentherapper, line 42: This is jumping ahead by one PR, but I think you'll want to match on `CertificateIdentities` rather than `Subjects` since we'd want to match on issuer as well.

mihaimaruseac (dismissed):

pkg/ct/monitor.go
linus-sun: going to address this in a follow-up PR

pkg/identity/identity.go
```diff
@@ -181,3 +194,145 @@ func MonitoredValuesExist(mvs MonitoredValues) bool {
...
+		return "", nil
+	}
+	return "", errors.New("certificate was neither x509 nor google_x509")
+}
+
+// OIDMatchesPolicy returns if a certificate contains both a given OID field and a matching value associated with that field
+// if true, it returns the OID extension and extension value that were matched on
+func OIDMatchesPolicy[Certificate *x509.Certificate | *google_x509.Certificate](certificate Certificate, oid asn1.ObjectIdentifier, extensionValues []string) (bool, asn1.ObjectIdentifier, string, error) {
+	switch cert := any(certificate).(type) {
+	case *x509.Certificate:
```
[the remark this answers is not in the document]
linus-sun: Without the switch statement, Golang won't detect that the Certificate generic constraint will always contain an Extensions field; the switch statement in OIDMatchesPolicy is extraneous and can be removed, but getExtension/getDeprecatedExtension both require the cases

pkg/identity/identity.go
```diff
@@ -181,3 +194,145 @@ func MonitoredValuesExist(mvs MonitoredValues) bool {
...
+			}
+			if len(rest) != 0 {
+				return "", fmt.Errorf("unmarshalling extension had rest for oid %v", oid)
+			}
+			return extValue, nil
+		}
+		return "", nil
+	case *google_x509.Certificate:
+		for _, ext := range cert.Extensions {
+			if !ext.Id.Equal((google_asn1.ObjectIdentifier)(oid)) {
```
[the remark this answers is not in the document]
linus-sun: same as above- would it be preferable to have two separate functions for getting extension between asn1/Google or is the switch statement preferable?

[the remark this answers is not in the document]
haydentherapper: This is good with me, this duplication is easy to spot then in case there's a later language feature to remove the duplication.

haydentherapper (approved):

PR #543, opened 2024-11-18
Change monitoring workflow configuration argument to optional

linus-sun:
#### Summary

This PR changes the `config` argument in the GitHub Actions reusable monitoring workflow to no longer be required, and sets the identity monitor portion of the workflow to only run if a `config` containing a desired set of identities to monitor has been passed in.

#### Release Note

Workflow `config` argument has been changed to optional

#### Documentation

N/A

### DIFF
--- a/cmd/rekor_monitor/main.go
+++ b/cmd/rekor_monitor/main.go
@@ -52,17 +52,10 @@ func main() {
 	configYamlInput := flag.String("config", "", "path to yaml configuration file containing identity monitor settings")
 	once := flag.Bool("once", true, "whether to run the monitor on a repeated interval or once")
 	serverURL := flag.String("url", publicRekorServerURL, "URL to the rekor server that is to be monitored")
+	logInfoFile := flag.String("file", logInfoFileName, "path to the initial log info checkpoint file to be read from")
 	interval := flag.Duration("interval", 5*time.Minute, "Length of interval between each periodical consistency check")
 	userAgentString := flag.String("user-agent", "", "details to include in the user agent string")
 	flag.Parse()
-
-	if *configFilePath == "" && *configYamlInput == "" {
-		log.Fatalf("empty configuration input")
-	}
-
-	if *configFilePath != "" && *configYamlInput != "" {
-		log.Fatalf("only input one of configuration file path or yaml input")
-	}
 
 	var config notifications.IdentityMonitorConfiguration
 
@@ -86,11 +79,6 @@ func main() {
 
 	if config.OutputIdentitiesFile == "" {
 		config.OutputIdentitiesFile = outputIdentitiesFileName
-	}
-
-	err := rekor.VerifyConsistencyCheckInputs(interval, &config.LogInfoFile, once)
-	if err != nil {
-		log.Fatal(err)
 	}
 
 	rekorClient, err := client.GetRekorClient(*serverURL, client.WithUserAgent(strings.TrimSpace(fmt.Sprintf("rekor-monitor/%s (%s; %s) %s", version.GetVersionInfo().GitVersion, runtime.GOOS, runtime.GOARCH, *userAgentString))))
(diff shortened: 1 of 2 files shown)


cmd/rekor_monitor/main.go
```diff
@@ -76,12 +69,16 @@ func main() {
 		if err := yaml.Unmarshal([]byte(configString), &config); err != nil {
 			log.Fatalf("error parsing identities: %v", err)
 		}
+
+		configProvided = true
```
haydentherapper, line 73: Do we need this? An uninitalized `config` should be sufficient for the checks below like `config.EndIndex == nil`

cmd/rekor_monitor/main.go
```diff
@@ -56,15 +56,8 @@ func main() {
 	userAgentString := flag.String("user-agent", "", "details to include in the user agent string")
 	flag.Parse()
 
-	if *configFilePath == "" && *configYamlInput == "" {
```
haydentherapper, line 59: I think something got dropped, I don't see logInfoFileName set as a default either.

cmd/rekor_monitor/main.go
```diff
@@ -76,12 +69,16 @@ func main() {
...
 
 	if *configYamlInput != "" {
 		if err := yaml.Unmarshal([]byte(*configYamlInput), &config); err != nil {
 			log.Fatalf("error parsing identities: %v", err)
 		}
+
+		configProvided = true
 	}
 
 	if config.OutputIdentitiesFile == "" {
```
haydentherapper, line 84: Below for `VerifyConsistencyCheckInputs`, given we provide defaults, do we need this check anymore? Just need to update `config.LogINfoFile` based on the default if it's not provided.

cmd/rekor_monitor/main.go
linus-sun: I believe that in the event that an uninitialized config is currently passed in, the workflow will still attempt to set `StartIndex` and `EndIndex` using the latest checkpoint and log checkpoint file and then attempt to run the identity monitor using the uninitialized config's values resulting in an error- if we would prefer to have checking for the config's indices stop the identity monitor from running I can refactor that in another PR after this one

cmd/rekor_monitor/main.go
linus-sun: sorry, mixed up with a future PR that I believe takes out logInfoFile from the config and resets it to a default argument- implementing that change in this PR as well

cmd/rekor_monitor/main.go
haydentherapper: > an uninitialized config will have the start and end indices set and then the identity monitor will run with nil arguments

I thought the defaults for start and end were [nil](https://github.com/sigstore/rekor-monitor/blob/114befbd5a9ab2639b80d592f7c777791495c192/pkg/notifications/notifications.go#L69-L70). So if the config is not initialized, then the default values of `nil` will be read in the code you linked, and then the monitor will initialize them with defaults.

cmd/rekor_monitor/main.go
linus-sun: reimplementing this as a default flag

cmd/rekor_monitor/main.go
linus-sun: updating logInfoFile and removing this part of the check

cmd/rekor_monitor/main.go
linus-sun: Yes - I guess what I'm thinking is that in the event that a config is not initialized, there are also no identities provided to monitor; is the identity monitor portion of the workflow then still necessary to run?

haydentherapper: Ah I see your question. What about if we gated calling `IdentitySearch` on if `MonitoredValues` is set? That was like the original implementation, where it would only run if a user configured a value to monitor for.

cmd/rekor_monitor/main.go
```diff
@@ -163,7 +162,6 @@ func main() {
 		}
```
haydentherapper, line 162: Can we remove the default of 0? Scrapping the entire log is hard to do without careful rate limiting on the client side. If no start index is provided and no checkpoint is provided, then just throw an error.

cmd/rekor_monitor/main.go
```diff
@@ -145,10 +138,16 @@ func main() {
 			}
 		}
 
+		err = rekor.RunConsistencyCheck(rekorClient, verifier, *logInfoFile)
```
haydentherapper, line 141: For a follow up - I noticed `RunConsistencyCheck` calls `GetLogInfo`, so we could reduce duplication by having `RunConsistencyCheck` return the log info struct.

cmd/rekor_monitor/main.go
```diff
@@ -88,7 +81,7 @@ func main() {
 		config.OutputIdentitiesFile = outputIdentitiesFileName
 	}
 
-	err := rekor.VerifyConsistencyCheckInputs(interval, &config.LogInfoFile, once)
+	err := rekor.VerifyConsistencyCheckInputs(interval, logInfoFile, once)
```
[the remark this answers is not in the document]
linus-sun: deleting this from `main.go` and removing the function in a follow-up PR

cmd/rekor_monitor/main.go
linus-sun: will work on this in a follow-up PR now

haydentherapper (approved):
Thanks!

PR #546, opened 2024-11-19
Refactor RunConsistencyCheck to return previous checkpoint 

linus-sun:
#### Summary

This PR refactors the RunConsistencyCheck() function to return the previously signed checkpoint for use in the IdentitySearch() function; this ensures that the identity monitor completes and fixes an error where the identity monitor attempted to run using the same checkpoint twice, resulting in an identical start and end index.

#### Release Note

NONE

#### Documentation

N/A

### DIFF
--- a/cmd/rekor_monitor/main.go
+++ b/cmd/rekor_monitor/main.go
@@ -28,7 +28,6 @@ import (
 	"github.com/sigstore/rekor-monitor/pkg/identity"
 	"github.com/sigstore/rekor-monitor/pkg/notifications"
 	"github.com/sigstore/rekor-monitor/pkg/rekor"
-	"github.com/sigstore/rekor-monitor/pkg/util/file"
 	"github.com/sigstore/rekor/pkg/client"
 	"github.com/sigstore/rekor/pkg/generated/models"
 	"github.com/sigstore/rekor/pkg/util"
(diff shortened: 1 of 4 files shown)


cmd/rekor_monitor/main.go
```diff
@@ -164,11 +157,13 @@ func main() {
 			return
 		}
 
-		// TODO: This should subsequently read from the identity metadata file to fetch the latest index.
-		_, err = rekor.IdentitySearch(*config.StartIndex, *config.EndIndex, rekorClient, monitoredValues, config.OutputIdentitiesFile, config.IdentityMetadataFile)
-		if err != nil {
-			fmt.Fprintf(os.Stderr, "failed to successfully complete identity search: %v", err)
-			return
+		if identity.MonitoredValuesExist(monitoredValues) {
+			// TODO: This should subsequently read from the identity metadata file to fetch the latest index.
```
haydentherapper, line 161: nit: we can remove this TODO since we said we'll stick with checkpoint indices

cmd/rekor_monitor/main.go
```diff
@@ -125,21 +124,15 @@ func main() {
...
-		logInfo, err = rekor.RunConsistencyCheck(rekorClient, verifier, *logInfoFile)
+		var prevCheckpoint *util.SignedCheckpoint
+		prevCheckpoint, logInfo, err = rekor.RunConsistencyCheck(rekorClient, verifier, *logInfoFile)
 		if err != nil {
 			fmt.Fprintf(os.Stderr, "error running consistency check: %v", err)
 			return
 		}
 
 		if config.StartIndex == nil {
 			if *logInfoFile != "" {
```
haydentherapper, line 135: Could we change this condition to be `if prevCheckpoint != nil`?

haydentherapper (changes requested):

haydentherapper (approved):

PR #547, opened 2024-11-19
YAML tag extensions

linus-sun:
#### Summary

Small PR to add YAML tagging to the OIDExtensions struct so that they are properly parsed by the rekor-monitor reusable monitoring workflow and added to the MonitoredValues struct from the ConfigMonitoredValues struct. As the MonitoredValues struct is no longer being parsed into, the YAML tags for it have been removed in favor of ConfigMonitoredValues.

#### Release Note

NONE

#### Documentation

N/A


mihaimaruseac (dismissed):
I think this makes sense, but you might want to still keep the ones in `identity.go` in case we might want to parse those from yaml?

haydentherapper (approved):

Issue #400, opened 2024-05-20
Add support for providing a URL to monitor


haydentherapper:
**Description**

The binary supports a `url` flag, defaulting to the production Rekor instance. The reusable workflow should take an optional parameter for the URL as well, so that the staging instance or private deployments can be monitored.

Horiodino:
@haydentherapper,  let me know if this issue is still open. Is there anything else I should be aware of? Thanks

haydentherapper:
Yep, it is! For this, we just need to add a configurable URL to the two reusable workflows and pass it to the binary with the `--url` flag.

PR #595, opened 2025-02-06
Added support for providing a custom URL to monitor

Horiodino:
#### Summary

Add option for using custom url . 
fixes: https://github.com/sigstore/rekor-monitor/issues/400

#### Documentation

the --url takes a url if its empty default is used.(please add it to the docks)

### DIFF
--- a/.github/workflows/ct_reusable_monitoring.yml
+++ b/.github/workflows/ct_reusable_monitoring.yml
@@ -26,6 +26,10 @@ on:
         description: 'multiline yaml of configuration settings for identity monitor run'
         required: false
         type: string
+      url:
+        description: 'Optional URL to pass to the monitor'
+        required: false
+        type: string
 
 permissions:
   contents: read
@@ -74,7 +78,13 @@ jobs:
         run: cat ${{ env.LOG_FILE }}
         # Skip on first run
         continue-on-error: true
-      - run: go run ./cmd/ct_monitor --config ${{ inputs.config }} --file ${{ env.LOG_FILE }} --once=${{ inputs.once }}
+      - run: |
+          go run ./cmd/ct_monitor \
+          --config ${{ inputs.config }} \
+          --file ${{ env.LOG_FILE }} \
+          --once=${{ inputs.once }} \
+          ${{ inputs.url && format('--url {0}', inputs.url) || '' }}
+
       - name: Upload checkpoint
         uses: actions/upload-artifact@65c4c4a1ddee5b72f698fdd19549f0f0fb45cf08 # v4.6.0
         with:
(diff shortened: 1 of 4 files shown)


Horiodino:
@haydentherapper I noticed that both Cli are using standard go pkg , not using the Cobra . I'm not sure if this was intentional, but I believe using Cobra would be a better approach since it's more maintainable and structured for CLI applications. Let me know your thoughts. If you agree, feel free to create an issue, and I'll handle the migration to Cobra. Otherwise no problem.

haydentherapper:
> @haydentherapper I noticed that both Cli are using standard go pkg , not using the Cobra . I'm not sure if this was intentional, but I believe using Cobra would be a better approach since it's more maintainable and structured for CLI applications. Let me know your thoughts. If you agree, feel free to create an issue, and I'll handle the migration to Cobra. Otherwise no problem.

Thanks, but given the simplicity of the codebase, I'd rather stick with the standard library unless we need a feature from a third-party library.

.github/workflows/ct_reusable_monitoring.yml
```diff
@@ -74,7 +78,13 @@ jobs:
         run: cat ${{ env.LOG_FILE }}
         # Skip on first run
         continue-on-error: true
-      - run: go run ./cmd/ct_monitor --config ${{ inputs.config }} --file ${{ env.LOG_FILE }} --once=${{ inputs.once }}
+      - run: |
+          go run ./cmd/ct_monitor \
+          --config ${{ inputs.config }} \
+          --file ${{ env.LOG_FILE }} \
+          --once=${{ inputs.once }} \
+          --url ${{ inputs.url }}
```
haydentherapper, line 86: Can we update this to be the same as the other command with the optional `--url` param?

.github/workflows/reusable_monitoring.yml
```diff
@@ -83,7 +87,13 @@ jobs:
...
         # Skip on first run
         continue-on-error: true
-      - run: go run ./cmd/rekor_monitor --file ${{ env.LOG_FILE }} --once=${{ inputs.once }} --user-agent "${{ format('{0}/{1}/{2}', needs.detect-workflow.outputs.repository, needs.detect-workflow.outputs.ref, github.run_id) }}" --config "${{ inputs.config }}"
+      - run: |
+          go run ./cmd/rekor_monitor \
+            --file ${{ env.LOG_FILE }} \
+            --once=${{ inputs.once }} \
+            --user-agent "${{ format('{0}/{1}/{2}', needs.detect-workflow.outputs.repository, needs.detect-workflow.outputs.ref, github.run_id) }}" \
+            --config "${{ inputs.config }}" \
+            ${{ inputs.url && format('--url {0}', inputs.url) || '' }}
```
haydentherapper, line 96: Cool! Have you verified this works as expected?

Horiodino: i wont be able to run test this locally . :sweat_smile:

haydentherapper (approved):
Thanks!

PR #665, opened 2025-05-19
Add Prometheus metrics for log consistency check

fghanmi:
#### Summary

This PR introduces Prometheus metrics support to `rekor_monitor` for tracking log consistency check operations. It adds two metrics:
- `log_index_verification_total`: Counter for the total number of log consistency check attempts.
- `log_index_verification_failure`: Counter for the number of failed log consistency check attempts.
Additionally, it introduces two new flags:
- `--run-as-monitor`: Enables Prometheus metrics collection and HTTP endpoint exposure.
- `--monitor-port`: Specifies the port for the Prometheus HTTP server (defaults to 9464).

This improves observability for users running Rekor monitor in long-lived environments or production settings, allowing for better alerting and dashboarding.


cmd/rekor_monitor/main.go
```diff
@@ -44,18 +51,52 @@ const (
...
+// Command-line flags that are parameters to the verifier job
+var (
+	configFilePath  = flag.String("config-file", "", "path to yaml configuration file containing identity monitor settings")
+	configYamlInput = flag.String("config", "", "path to yaml configuration file containing identity monitor settings")
+	once            = flag.Bool("once", true, "whether to run the monitor on a repeated interval or once")
+	serverURL       = flag.String("url", publicRekorServerURL, "URL to the rekor server that is to be monitored")
+	logInfoFile     = flag.String("file", logInfoFileName, "path to the initial log info checkpoint file to be read from")
+	interval        = flag.Duration("interval", 5*time.Minute, "Length of interval between each periodical consistency check")
+	userAgentString = flag.String("user-agent", "", "details to include in the user agent string")
+	runAsMonitor    = flag.Bool("run-as-monitor", false, "Enable this option to expose log consistency check results as Prometheus metrics on /metrics")
```
haydentherapper, line 63: Can we lump this together with `--once`? If `once` is false, then we can assume it's running as a long-lived process and we'll enable prometheus.

cmd/rekor_monitor/main.go
```diff
@@ -44,18 +51,52 @@ const (
...
+
+func handleError(msg string, err error) {
+	errWrap := errors.Join(errors.New(msg), err)
+	fmt.Fprint(os.Stderr, errWrap, "\n")
+
+	if runAsMonitor != nil && *runAsMonitor {
+		errStr := errWrap.Error()
+		// These specific messages are expected in normal operation and are not treated as consistency check failures.
+		// Therefore, they are excluded from Prometheus failure metrics.
+		if strings.Contains(errStr, "consistency proofs can not be computed starting from an empty log") ||
```
haydentherapper, line 86: What is this error from? I wouldn't expect a monitor to be running against an empty log.

cmd/rekor_monitor/main.go
```diff
@@ -82,6 +123,27 @@ func main() {
 		config.OutputIdentitiesFile = outputIdentitiesFileName
 	}
 
+	if runAsMonitor != nil && *runAsMonitor {
+		go func() {
```
haydentherapper, line 127: The main script has gotten a bit unwieldy unfortunately. Can we split the initialization of metrics and starting up the server into its own file, also so that we can add tests?

https://github.com/sigstore/rekor-tiles/blob/main/pkg/server/metrics.go is a decent example, just ignoring the bits about gRPC.

This also gives an example of gracefully shutting down the metrics server when sent a SIGINT/SIGTERM in case this is running on K8s.

It'd also be nice to not use promauto and initialize a new registry, in case this gets refactored into a library to avoid any collisions of prom registries.

cmd/rekor_monitor/main.go
```diff
@@ -126,6 +188,9 @@ func main() {
 
 	// To get an immediate first tick, for-select is at the end of the loop
 	for {
+		if runAsMonitor != nil && *runAsMonitor {
```
haydentherapper, line 191: I think we can unconditionally increment metrics since these are global variables and always initialized.

cmd/rekor_monitor/main.go
```diff
@@ -134,41 +199,51 @@ func main() {
 		var prevCheckpoint *util.SignedCheckpoint
 		prevCheckpoint, logInfo, err = rekor.RunConsistencyCheck(rekorClient, verifier, *logInfoFile)
 		if err != nil {
-			fmt.Fprintf(os.Stderr, "error running consistency check: %v", err)
-			return
+			handleError("error running consistency check", err)
+			if runAsMonitor != nil && *runAsMonitor {
+				goto waitForTick
```
haydentherapper, line 204: I think the goto needs to be before where we check `if *once` so that it exits early. Also feel free to refactor anything here if it would help with the control flow, this is going to get cleaned up soon!

haydentherapper (changes requested):
Sorry about the delay, thanks so much for making these changes!

cmd/rekor_monitor/main.go
fghanmi: This error comes from rekor verify package: https://github.com/sigstore/rekor/blob/main/pkg/verify/verify.go#L45

When I run the monitor on an empty rekor log, this is the output that I get:
```
error running consistency check: failed to verify previous checkpoint: failed to verify log consistency: consistency proofs can not be computed starting from an empty log
```

cmd/rekor_monitor/main.go
fghanmi: @haydentherapper all `goto waitForTick` are indeed before `if *once` or I am not seeing it well ?

pkg/server/metrics.go
```diff
@@ -0,0 +1,118 @@
...
+	signal.Notify(signalChan, os.Interrupt, syscall.SIGTERM)
+}
+
+// InitRegistryForTesting resets the registry for test isolation.
+// This is only used in tests to ensure a clean state.
+func InitRegistryForTesting() {
+	registry = prometheus.NewRegistry()
+	registry.MustRegister(logIndexVerificationTotal, logIndexVerificationFailure)
+}
+
```
ret2libc, line 57: would it be better to put this elsewhere to separate regular code from testing code?

pkg/server/metrics.go
```diff
@@ -0,0 +1,118 @@
...
+		Name: "log_index_verification_total",
+		Help: "Total number of log consistency check attempts.",
+	})
+	logIndexVerificationFailure = prometheus.NewCounter(prometheus.CounterOpts{
+		Name: "log_index_verification_failure",
+		Help: "Total number of failed log consistency check attempts.",
+	})
+
+	signalChan = make(chan os.Signal, 1)
+)
```
ret2libc, line 44: i'd split this similarly to how it's done in https://github.com/sigstore/rekor-tiles/blob/main/internal/server/metrics.go#L52C6-L52C16

ret2libc: maybe doing that would also help in better test/production code separation (e.g. https://github.com/sigstore/rekor-tiles/blob/main/internal/server/metrics_test.go#L25 )

pkg/server/metrics.go
```diff
@@ -0,0 +1,118 @@
...
+		ReadTimeout:       10 * time.Second,
+		WriteTimeout:      10 * time.Second,
+		IdleTimeout:       120 * time.Second,
+		ReadHeaderTimeout: 5 * time.Second,
+	}
+
+	// Start server in a goroutine
+	go func() {
+		if err := server.ListenAndServe(); err != nil && err != http.ErrServerClosed {
+			log.Fatalf("Failed to start Prometheus metrics server: %v", err)
```
ret2libc, line 101: should this really be a fatal? Can we return an error and let the caller decide what to do with it?

internal/cmd/common.go
```diff
@@ -59,20 +62,42 @@ type MonitorLoopParams struct {
...
+}
+
 type Checkpoint interface{}
 type LogInfo interface{}
 
 // ParseMonitorFlags parses command-line flags and returns a MonitorFlags struct
 func ParseMonitorFlags(defaultServerURL, defaultTUFRepository string, baseUserAgentName string) *MonitorFlags {
 	configFilePath := flag.String("config-file", "", "path to yaml configuration file containing identity monitor settings")
 	configYamlInput := flag.String("config", "", "string with YAML configuration containing identity monitor settings")
-	once := flag.Bool("once", true, "whether to run the monitor on a repeated interval or once")
```
ret2libc, line 69: you can keep this here instead of https://github.com/sigstore/rekor-monitor/pull/665/files#diff-83a2daadb61b967d7cd2c064c50ee421a2b730cc93a98c1bad36089f863d446bR65-R67 and use `params.Once` in the `MonitorLoop` function

cmd/ct_monitor/main.go
```diff
@@ -59,6 +60,11 @@ func main() {
 		logInfoFileName := fmt.Sprintf("%s.txt", logInfoFileName)
 		flags.LogInfoFile = logInfoFileName
 	}
+	if !flags.Once {
+		if err := server.StartMetricsServer(flags.MonitorPort); err != nil {
+			log.Fatalf("Failed to start Prometheus metrics server: %v", err)
+		}
+	}
```
ret2libc, line 67: this seems to be a common operation to all monitors, we might just move it in the `MonitorLoop` function

internal/cmd/common.go
```diff
@@ -252,12 +288,16 @@ func MonitorLoop(params MonitorLoopParams) {
...
 		select {
 		case <-ticker.C:
 			continue
 		case <-ctx.Done():
 			fmt.Fprintf(os.Stderr, "Shutting down gracefully...")
 			return
+		case <-server.GetSignalChan():
+			fmt.Fprintf(os.Stderr, "received signal, exiting")
+			return
 		}
```
ret2libc, line 301: can we handle the shutdown better if we pass the `ctx` to the metrics server?

internal/cmd/common.go
```diff
@@ -196,12 +220,19 @@ func MonitorLoop(loopLogic MonitorLogic) {
...
+					}
 				}
 
 				foundEntries, failedEntries, err := loopLogic.IdentitySearch(ctx, config, loopLogic.MonitoredValues())
 				if err != nil {
-					fmt.Fprintf(os.Stderr, "failed to successfully complete identity search: %v", err)
+					fmt.Fprintf(os.Stderr, "failed to successfully complete identity search: %v\n", err)
+					if wErr := loopLogic.WriteCheckpoint(prevCheckpoint, curCheckpoint); wErr != nil {
+						fmt.Fprintf(os.Stderr, "failed to write checkpoint: %v\n", wErr)
+					}
```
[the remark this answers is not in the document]
fghanmi: I thought to keep the behavior of the previous `common.go` to have this test succeed: https://github.com/sigstore/rekor-monitor/blob/8c117c7815cea3f44c424ea2ab31815c63e41a76/internal/cmd/common_test.go#L760

Old `common.go` always persist the checkpoint, even if identity search fails. 
But, I will update the test value and only persist a “clean” state as you suggest here.

internal/cmd/common.go
```diff
@@ -167,17 +173,31 @@ func MonitorLoop(loopLogic MonitorLogic) {
...
-			return
+			fmt.Fprintf(os.Stderr, "error running consistency check: %v\n", err)
+			if loopLogic.Once() {
+				return
+			} else {
+				if !strings.Contains(err.Error(), "consistency proofs can not be computed starting from an empty log") {
+					server.IncLogIndexVerificationFailure()
+				}
+				goto waitForTick
+			}
```
ret2libc, line 200: I think this is similar to https://github.com/sigstore/rekor-monitor/pull/665#discussion_r2319463775 . We should probably handle this properly elsewhere? Is this just a way to workaround https://github.com/sigstore/rekor-monitor/pull/702 ? If that's the case, I think this shouldn't be needed if we merge that PR. Right?

ret2libc:
Nicely done! Thanks for the work!

mihaimaruseac (approved):

PR #671, opened 2025-05-23
Update documentation for GetLogInfo

haydentherapper:
We now fetch the latest checkpoint, not the "stable" (every 5 min) checkpoint.


haydentherapper:
A log could split-view a monitor - the only solution is for a monitor to request that the latest checkpoint be co-signed.

For this change, this was when we were thinking that the monitor itself could also be a witness, fetching a "stable" (periodic) checkpoint published from Rekor v1 since Rekor publishes checkpoints too frequently for two witnesses to have the same state. In practice, this was never actually used.

PR #688, opened 2025-06-25
Add support for CT log entries with Precertificates

facutuesca:
#### Summary

Monitoring identities in the CT log currently segfaults due to the log having entries with [Precertificates](https://www.rfc-editor.org/rfc/rfc6962#section-3.1). Since the monitor [assumes](https://github.com/sigstore/rekor-monitor/blob/dbbfb0f5466c0f69f40995b81d7a2f6fe7b28950/pkg/ct/monitor.go#L38) that entries always have a non-nil [`X509Cert`](https://github.com/google/certificate-transparency-go/blob/711a92c67f051aa8567ad723446fc54fc14e4662/types.go#L229-L230), this causes a segfault when entries contain a Precertificate instead.

This PR fixes this by using the (extracted) Precertificate if `LogEntry.X509Cert` is found to be nil. It also fails gracefully if it ever happens that both are nil, instead of segfaulting.

cc @haydentherapper

### DIFF
--- a/pkg/ct/monitor_test.go
+++ b/pkg/ct/monitor_test.go
@@ -38,7 +38,8 @@ func TestScanEntryCertSubject(t *testing.T) {
 	testCases := map[string]struct {
 		inputEntry    ct.LogEntry
 		inputSubjects []identity.CertificateIdentity
-		expected      []identity.LogEntry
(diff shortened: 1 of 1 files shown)


haydentherapper (approved):
Nice, good find

Issue #51, opened 2022-08-27
Consistency: Verify checkpoint with TUF metadata


haydentherapper:
**Description**

Currently the Rekor public key, needed to verify checkpoint signatures, is fetched from Rekor. This is not ideal, and we should use the public key in the TUF targets to verify checkpoints. This will mean we need to bundle the TUF root.

linus-sun:
This also applies [here](https://github.com/sigstore/rekor-monitor/blob/48375db9da425b4d9985ca13cb9f854b4f59439d/pkg/ct/consistency.go#L43)- the CTFE public key should be fetched from TUF.

facutuesca:
> This will mean we need to bundle the TUF root.

@haydentherapper if we were to get the TUF targets using `sigstore-go`'s TUF [client](https://github.com/sigstore/sigstore-go/tree/main/pkg/tuf), do we need to bundle the roots in this repo? I ask since `sigstore-go` already [bundles them](https://github.com/sigstore/sigstore-go/tree/main/pkg/tuf/repository) and you can create a client with either Prod or Staging roots:

https://github.com/sigstore/sigstore-go/blob/7212b45515a1387f7c8dd147747838355ce72427/pkg/tuf/options.go#L138-L141

haydentherapper:
@facutuesca No, for the public instance, we can leverage sigstore-go directly and don't need to bundle the roots. We can add flags to provide your own root and and TUF repo and/or public key if you're not monitoring the public instance.

facutuesca:
@haydentherapper I think this one can be closed

PR #695, opened 2025-07-07
Fetch public keys using TUF

facutuesca:
#### Summary

Part of https://github.com/sigstore/rekor-monitor/issues/51. 

This changes both `rekor_monitor` and `ct_monitor` to fetch the public keys using Sigstore's [trust root](https://tuf-repo-cdn.sigstore.dev/targets/f44a1b88128e55ebfb62189becbc0fa48d4ec9915c65ac54ba0e46a008b12d5b.trusted_root.json) (via `sigstore-go`). 

For Rekor, we extract the `logId` from the latest checkpoint and use it to select a matching `TransparencyLog` instance from the Trusted Root. For CT, we hardcode the `logId` for `https://ctfe.sigstore.dev/2022` and use it to select the corresponding `TransparencyLog` instance from the trusted root.

Something I couldn't figure out: is there a way to avoid hardcoding the `logId` for `https://ctfe.sigstore.dev/2022`?
From what I could see, the latest CT checkpoint doesn't have the information needed to select a transparency log from the trusted root.

cc @haydentherapper


pkg/ct/consistency.go
```diff
@@ -22,34 +22,53 @@ import (
...
 	"github.com/transparency-dev/merkle/proof"
 	"github.com/transparency-dev/merkle/rfc6962"
 )
 
 const (
-	ctfe2022PubKey = `-----BEGIN PUBLIC KEY-----
-MFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAEiPSlFi0CmFTfEjCUqF9HuCEcYXNK
-AaYalIJmBZ8yyezPjTqhxrKBpMnaocVtLJBI1eM3uXnQzQGAJdJ4gs9Fyw==
------END PUBLIC KEY-----`
+	ctfe2022KeyID = "dd3d306ac6c7113263191e1c99673702a24a5eb8de3cadff878a72802f29ee8e"
```
haydentherapper, line 32: As you found, there's no good way to do this, as the checkpoint format for CT logs (e.g. https://ctfe.sigstore.dev/2022/ct/v1/get-sth) is quite different. With [static ct](https://github.com/C2SP/C2SP/blob/main/static-ct-api.md), this will change. This is good for now!

haydentherapper (commented):
Looks great! Just a failure in the e2e test to look into.

haydentherapper (approved):

PR #705, opened 2025-08-01
Initial support for Rekor v2

facutuesca:
## Summary
Add initial support for monitoring Rekor v2 logs. Currently we only check the consistency of the log (no identity monitoring yet).

Test by running:
```shell
make build
./rekor_monitor -url https://log2025-alpha1.rekor.sigstage.dev -tuf-repository staging -once=false -interval=1s
```

## Details
- A new `tuf-repository` flag is added to the `rekor_monitor` CLI. It currently accepts either `default` or `staging` as values.
- We determine which version the server uses (`v1` or `v2`) by fetching the `SigningConfig` from the TUF repo and matching against the server URL (passed as a CLI flag with `-url`)
- The default filename of the checkpoint file is now different depending on the version 
   - (`logInfo.txt` -> `logInfo.v1.txt` or `logInfo.v2.txt`)
- The `pkg/rekor` package is moved to `pkg/rekor/v1`, and a new `pkg/rekor/v2` package is added. 
  - Currently, there's no logic shared between `v1` and `v2`, since the logic for consistency checks is different. When we add identity monitoring to v2, the shared logic (for identities) can be moved to the parent `rekor` package.
- We get the server's public key by matching the checkpoint key ID with the entries in the Trusted Root
- For fetching the public key from the Trusted Root, we need the checkpoint Key ID of the server. But fetching the checkpoint requires the public key (for most of the existing APIs). In order to get around this, we fetch the checkpoint manually without verifying it (at first), in order to get the key ID we need to get the public key.
  - See the comments for  `GetCheckpointKeyIDUnverified` for more details.
- We support sharding: we get all shards from the `SigningConfig` and perform the consistency check against the appropriate shard.

cc @haydentherapper


pkg/rekor/v2/verifier.go
```diff
@@ -0,0 +1,180 @@
...
+			return nil, fmt.Errorf("failed to build consistency proof: %v", err)
+		}
+
+		err = proof.VerifyConsistency(rfc6962.DefaultHasher, prevCheckpoint.Size, newCheckpoint.Size, consistencyProof, prevCheckpoint.Hash, newCheckpoint.Hash)
+		if err != nil {
+			return nil, fmt.Errorf("consistency check failed: %v", err)
+		}
+
+		fmt.Fprintf(os.Stderr, "Root hash consistency verified - Current Size: %d Root Hash: %s - Previous Size: %d Root Hash %s\n",
+			newCheckpoint.Size, hex.EncodeToString(newCheckpoint.Hash), prevCheckpoint.Size, hex.EncodeToString(prevCheckpoint.Hash))
```
ret2libc, line 167: i'd make this a separate function like what's done in v1 with `verifyCheckpointConsistency`.

pkg/rekor/v2/verifier.go
```diff
@@ -0,0 +1,180 @@
...
+
+	// Write if there was no stored checkpoint or the origin/sizes differ
+	if prevCheckpoint == nil || prevCheckpoint.Origin != latestActiveCheckpoint.Origin || prevCheckpoint.Size != latestActiveCheckpoint.Size {
+		if err := file.WriteCheckpointRekorV2(latestActiveCheckpoint, logInfoFile); err != nil {
+			// TODO: Once the consistency check and identity search are split into separate tasks, this should hard fail.
+			// Temporarily skipping this to allow this job to succeed, remediating the issue noted here: https://github.com/sigstore/rekor-monitor/issues/271
+			fmt.Fprintf(os.Stderr, "failed to write checkpoint: %v", err)
+		}
+	}
+
```
ret2libc, line 178: should we have a call to `DeleteOldCheckpoints` similar to v1?

cmd/rekor_monitor/main.go
```diff
@@ -102,8 +111,52 @@ func main() {
 	if err != nil {
 		log.Fatal(err)
 	}
+	// TODO: Use root.GetSigningConfig once https://github.com/sigstore/sigstore-go/pull/506 is merged
```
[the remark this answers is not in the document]
facutuesca: yeah, it was released in `sigstore-go 1.1.1`. I update it in the follow-up PR here: https://github.com/trail-of-forks/rekor-monitor/pull/1 (since the API also changes a bit), so I'll fix the `GetSigningConfig` use there

pkg/rekor/v2/verifier.go
facutuesca: not sure I agree. This function is already called `RunConsistencyCheck`, so it's clear it's about checking the consistency, and it's not too big (~70 lines long). 

Moving the meat of it to a separate `verifyCheckpointConsistency` does not add much in terms of readability IMHO.

pkg/rekor/v2/verifier.go
facutuesca: not necessary, since we'll switch to only storing the latest checkpoint anyway

pkg/util/file/file.go
```diff
@@ -112,6 +142,23 @@ func WriteCheckpoint(checkpoint *util.SignedCheckpoint, logInfoFile string) erro
...
+	file, err := os.OpenFile(logInfoFile, os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0644)
+	if err != nil {
+		return fmt.Errorf("error opening file: %w", err)
+	}
+	defer file.Close()
+	// Replace newlines to flatten checkpoint to single line
+	if _, err := fmt.Fprintf(file, "%s\n", strings.ReplaceAll(string(s), "\n", "\\n")); err != nil {
+		return fmt.Errorf("failed to write to file: %w", err)
+	}
+	return nil
```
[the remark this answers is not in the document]
facutuesca: fixed

cmd/rekor_monitor/main.go
```diff
@@ -102,8 +111,52 @@ func main() {
...
+		}
+		mainLoopV2(*once, *logInfoFile, *interval, rekorShards, activeShardOrigin)
+		return
+	default:
+		log.Fatalf("Unsupported server version %v, only '1' and '2' are supported", rekorVersion)
+	}
+}
+
+func mainLoopV1(serverURL string, once bool, logInfoFile string, interval time.Duration, userAgentString string, config notifications.IdentityMonitorConfiguration, trustedRoot *root.TrustedRoot) {
+	rekorClient, err := client.GetRekorClient(serverURL, client.WithUserAgent(strings.TrimSpace(fmt.Sprintf("rekor-monitor/%s (%s; %s) %s", version.GetVersionInfo().GitVersion, runtime.GOOS, runtime.GOARCH, userAgentString))))
```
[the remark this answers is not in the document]
facutuesca: fixed

ret2libc (approved):

PR #725, opened 2025-08-20
Refactor IdentitySearch logic between ct and rekor

ret2libc:
#### Summary
`ct_monitor` was not printing the identity matched, similar to what `rekor_monitor` is doing. This change refactors the `IdentitySearch` functions for both ct/rekor to share the same logic, so that also `ct_monitor` prints the matched identities.

#### Release Note
* Make `ct_monitor` print matched identities similar to `rekor_monitor`.


mihaimaruseac (dismissed):

mihaimaruseac (approved):

pkg/util/identity/identity.go
facutuesca, line 1: is there a reason to create a new `pkg/util/identity` pkg for this? Could `ProcessMatchedEntries` be part of `pkg/identity` instead?

facutuesca: Ah yeah that makes sense. Looking at `ProcessMatchedEntries`, it seems like it does two (independent) things:
- Write the matched entries to the output files
- Convert the matched entries to a list of `MonitoredIdentity`

Would it make sense to extract the first part to a function in `pkg/util/file`, and don't have a separate function for the second? (since it's only two lines of code)


Something like:
```go
// util/file/file.go
func WriteMatchedIdentityEntries(...) {
    ....
}

// ct/monitor.go
func IdentitySearch(...) {
    ....
    WriteMatchedIdentityEntries(...)
   
    identities := identity.CreateIdentitiesList(monitoredValues)
    return  identity.CreateMonitoredIdentities(matchedEntries, identities), nil
}

// rekor/v1/identity.go
func IdentitySearch(...) {
    ....
    WriteMatchedIdentityEntries(...)
   
    identities := identity.CreateIdentitiesList(monitoredValues)
    return  identity.CreateMonitoredIdentities(matchedEntries, identities), nil
}
```

PR #726, opened 2025-08-21
Deduplicate ct and rekor monitors

ret2libc:
#### Summary
Until now, ct-monitor and rekor-monitor were duplicating a lot of code to perform the loop, check various arguments, print data, etc. The result was that changes implemented in one implementation did not get always ported to the other. These changes refactor that code in a `common.go` file that implements the main loop over the logs (CT or Rekor).

### DIFF
--- a/cmd/common.go
+++ b/cmd/common.go
@@ -1,0 +1,224 @@
+//
+// Copyright 2025 The Sigstore Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package cmd
+
+import (
+	"context"
+	"flag"
+	"fmt"
+	"os"
+	"os/signal"
+	"runtime"
+	"strings"
+	"syscall"
+	"time"
+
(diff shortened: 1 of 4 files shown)


mihaimaruseac (approved):

cmd/common.go
facutuesca, line 1: I _think_ the usual thing to do here is to put the shared files (`common.go`) that are not meant to be externally accessed in a separate top-level dir called `internal`, e.g.: https://github.com/sigstore/rekor-tiles/tree/main/internal

cmd/common.go
```diff
@@ -0,0 +1,224 @@
...
+	}
+	for _, fp := range monitoredValues.Fingerprints {
+		fmt.Printf("Monitoring fingerprint %s\n", fp)
+	}
+	for _, sub := range monitoredValues.Subjects {
+		fmt.Printf("Monitoring subject %s\n", sub)
+	}
+}
+
+func LoopLogs(params LoopLogsParams) {
```
facutuesca, line 153: nit: maybe `MonitorLoop` would be clearer? (and `MonitorLoopParams`)

cmd/rekor_monitor/main.go
```diff
@@ -173,116 +127,71 @@ func mainLoopV1(serverURL string, once bool, logInfoFile string, interval time.D
...
+				prevCheckpoint = prev
+			}
+			var curCheckpoint cmd.Checkpoint
+			if cur != nil {
+				curCheckpoint = cur
+			}
+			return prevCheckpoint, curCheckpoint, nil
+		},
+		GetStartIndexFn: func(prev, cur cmd.Checkpoint) *int {
+			checkpointStartIndex := rekor_v1.GetCheckpointIndex(cur.(*models.LogInfo), prev.(*util.SignedCheckpoint))
```
facutuesca, line 152: I find this a bit confusing. Both `prev` and `cur` are `cmd.Checkpoint`, but they are not both checkpoints (`prev` is a checkpoint, `cur` is a `LogInfo`)

PR #729, opened 2025-08-21
Handle sharding of Rekor v2 log while monitor runs

facutuesca:
#### Summary

- The first commit adds logic to the beginning of each v2 monitor iteration, to handle the case where a new shard has been added while the monitor is running. We check if the `SigningConfig` target has changed, and if it has we update the list of Rekor shards
- The second commit fixes a bug where the monitor would fail if the stored checkpoint was from a previous shard that is no longer valid. Since the `root.SelectServices` function only returns the currently valid shards, this would not include the shard that the stored checkpoint belonged to.
The solution is to replace the `SelectServices` call with a simplified version of its logic. We only need to filter the shards so that they match the v2 version, and sort them from newest to oldest. That way we get all the existing shards (valid and invalid) that we need for monitoring.

### DIFF
--- a/pkg/rekor/v2/client.go
+++ b/pkg/rekor/v2/client.go
@@ -57,11 +57,10 @@ func filterV2Shards(rekorServices []root.Service) []root.Service {
 	slices.SortFunc(sortedServices, func(i, j root.Service) int {
 		return i.ValidityPeriodStart.Compare(j.ValidityPeriodStart)
 	})
-	slices.Reverse(sortedServices)
 
 	var rekorV2Services []root.Service
 	now := time.Now()
-	for _, s := range sortedServices {
+	for _, s := range slices.Backward(sortedServices) {
 		if s.MajorAPIVersion == 2 && !s.ValidityPeriodStart.IsZero() && s.ValidityPeriodStart.Before(now) {
 			rekorV2Services = append(rekorV2Services, s)
 		}


pkg/rekor/v2/client.go
```diff
@@ -18,29 +18,105 @@ import (
...
+	for _, s := range sortedServices {
+		if s.MajorAPIVersion == 2 {
+			rekorV2Services = append(rekorV2Services, s)
+		}
+	}
+
+	return rekorV2Services
+}
+
+func ShardsNeedUpdating(currentShards map[string]ShardInfo, newSigningConfig *root.SigningConfig) (bool, error) {
```
ret2libc, line 70: could we run `GetRekorShards` on the new signing config and do a `reflect.DeepEqual(old, new)` instead? Or is this different in some subtle way?

ret2libc (commented):
A few points:
- is it fine if the user provides a server-URL that refers to an invalid shard? 
- can we simplify the `main()` logic so that rekorShards/activeShardOrigin are always computed inside the `RunConsistencyCheckFn` function? So `mainLoopV2` can just be `mainLoopV2(flags, config, tufClient)`

pkg/rekor/v2/client.go
```diff
@@ -50,7 +126,7 @@ func GetRekorShards(ctx context.Context, trustedRoot *root.TrustedRoot, rekorSer
 			return nil, "", err
 		}
 
-		// The services in rekorServices are ordered from newest to oldest,
+		// The services in rekorV2Services are ordered from newest to oldest,
```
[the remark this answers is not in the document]
facutuesca: Yeah this is a good point. It could be that the the server has no active shards (the latest shard is no longer valid), or that the latest shard is not _yet_ valid (will become valid in the future).

For the first case, I think it's fine, the shard not being active means that nothing new is being written to it, but we can still check the consistency with regards to a previously-stored checkpoint.

For the second case, yeah this is a good point. I believe we don't care about "future active" shards, since we can only start monitoring them once they're active, so I modified the `filterV2Shards` function to only return shards that are or have been active.

pkg/rekor/v2/client.go
facutuesca: `GetRekorShards` does some extra network requests in order to populate the `ShardInfo` struct. It creates a new client per shard, fetches the latest checkpoint for each, and creates a verifier for each.

Since `ShardsNeedUpdating` will run on every iteration, I made it so that it only compares the metadata in `SigningConfig` without any extra network requests.

facutuesca:
> A few points:
> 
> * is it fine if the user provides a server-URL that refers to an invalid shard?

Yeah, actually for v2 the URL is just used during the common `main()` logic, to see if we are dealing with a v1 or v2 server. The `mainLoopV2` function does not use it after that.

> * can we simplify the `main()` logic so that rekorShards/activeShardOrigin are always computed inside the `RunConsistencyCheckFn` function? So `mainLoopV2` can just be `mainLoopV2(flags, config, tufClient)`

Mmm maybe. I need to check once we adapt the identity monitoring logic for v2 to the new `MonitorLoop` struct, since it also needs access to the updated shards.

pkg/rekor/v2/client.go
```diff
@@ -18,29 +18,107 @@ import (
...
 
-func GetRekorShards(ctx context.Context, trustedRoot *root.TrustedRoot, rekorServices []root.Service, userAgent string) (map[string]ShardInfo, string, error) {
-	rekorServices, err := root.SelectServices(rekorServices, root.ServiceConfiguration{Selector: prototrustroot.ServiceSelector_ALL}, []uint32{2}, time.Now())
+func RefreshSigningConfig(tufClient *tuf.Client) (*root.SigningConfig, error) {
+	err := tufClient.Refresh()
+	if err != nil {
+		return nil, fmt.Errorf("error refreshing TUF client: %v", err)
+	}
+
+	signingConfig, err := root.GetSigningConfig(tufClient)
```
ret2libc, line 43: we should also update the main.go file to use `root.GetSigningConfig(tufClient)` (but ok to do in another PR).

PR #730, opened 2025-08-25
cmd/rekor_monitor: accept custom TUF

ret2libc:
#### Summary
`rekor-monitor` uses TUF metadata to know which Rekor version is used, to retrieve the verifier, trusted_root, etc. If we want to test `rekor-monitor` on local deployments or non standard ones, we need a way to specify a custom TUF.

#### Release Note
* Add support for custom TUF repository in `rekor-monitor`


facutuesca (approved):
👍

PR #736, opened 2025-08-26
pkg/notifications: make Notifications more customazible

ret2libc:
#### Summary
Make the Notifications system more flexible, so that it can be used for both ct-monitor and rekor-monitor and prepare for reporting not only matched identities, but also other kind of data (e.g. failed entries).

### DIFF
--- a/pkg/notifications/github_issues.go
+++ b/pkg/notifications/github_issues.go
@@ -16,15 +16,14 @@ package notifications
 
 import (
 	"context"
+	"fmt"
 	"strings"
 
 	"github.com/google/go-github/v65/github"
-	"github.com/sigstore/rekor-monitor/pkg/identity"
 )
 
 var (
-	notificationPlatformGitHubIssueBodyHeaderText = "Rekor-monitor found the following pairs of monitored identities and matching log entries: "
-	notificationPlatformGitHubIssueLabels         = []string{"rekor-monitor", "automatically generated"}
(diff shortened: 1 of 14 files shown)


mihaimaruseac (dismissed):

pkg/notifications/github_issues.go
```diff
@@ -41,31 +40,34 @@ type GitHubIssueInput struct {
 	GitHubClient *github.Client `yaml:"githubClient"`
 }
 
-func generateGitHubIssueBody(monitoredIdentities []identity.MonitoredIdentity) (string, error) {
-	header := notificationPlatformGitHubIssueBodyHeaderText
-	body, err := identity.PrintMonitoredIdentities(monitoredIdentities)
+// generateGitHubIssueBody generates a GitHub issue body for generic notification data
+func generateGitHubIssueBody(data NotificationData) (string, error) {
+	// Use custom header if available in metadata, otherwise use default
```
ret2libc, line 45:
```suggestion
```

pkg/notifications/github_issues.go
```diff
@@ -41,31 +40,34 @@ type GitHubIssueInput struct {
...
+	header := fmt.Sprintf(notificationPlatformGitHubIssueBodyHeaderText, data.Context.MonitorType)
+
+	// Try to use the payload as a NotificationBodyConverter
```
ret2libc, line 48:
```suggestion
```

Issue #386, opened 2024-05-03
Write to logInfo.txt only after identities.txt have been correctly updated


xiaokangwang:
Currently, when running verifier on a local system, it will resume from the point recorded in logInfo.txt. However, it overrides logInfo.txt with the most recent data point before entries between `Previous Size` and `Current Size` is scanned for `monitored-values`. By shutting down the monitor while it is running, the identities.txt could have missing entries that could potentially break the monitoring process.

haydentherapper:
This sounds good to me. 

Overall this needs to be reworked and each of these should be able to fail independent of one another without causing the verifier to stop.

ret2libc:
Slightly related to this, I think there's also some inconsistencies/lack of clarity around start/end indexes for monitoring identities. If you don't specify any `startIndex`/`endIndex` in the config, those will be computed based on the checkpoints, thus all entries between the two checkpoints will be analyzed (or at least should).

However, if you do specify a `startIndex`/`endIndex`, then the checkpoints and the identity monitors will go in different directions, with the identity monitor always checking the same set of log entries (because `once` is ignored when you specify an `endIndex` and the loop always terminates after the first run).

What was the purpose of the start/end indexes configuration? Should they really be configurable? cc @haydentherapper

haydentherapper:
Providing only `startIndex` was if the monitor needed to be restarted and there was no local checkpoint. Providing both `startIndex` and `endIndex` was to "backfill" monitor in case there was some period that was skipped over. I can't think of a use case for just an `endIndex`. I do agree this can be cleaned up though.

PR #740, opened 2025-08-27
Split the consistency check and the checkpoint writing

ret2libc:
#### Summary
Before this patch, the consistency check function was also writing the new checkpoint to file, before searching for matching identities. If something would happen during the identity matching, by the time you re-run the monitor you'd start from the new checkpoint, rather than the previous one, skipping the entries that were part of the last checkpoint "diff".

This patch splits the checkpoint consistency check and the storing of the new checkpoint, so that it is written only after all identities have been searched. If an error happens during the identity search, the new checkpoint would not be written to file and the next monitor run would start again from the old checkpoint (at the expense of extra time required to re-analyze all the entries).

Fixes https://github.com/sigstore/rekor-monitor/issues/386.

#### Release Note
* Ensure identities are never skipped, even if notifications fail, by updating the current checkpoint only after having scanned all entries from the previous to the current checkpoint.

### DIFF
--- a/cmd/rekor_monitor/main.go
+++ b/cmd/rekor_monitor/main.go
@@ -32,6 +32,7 @@ import (
 	"github.com/sigstore/rekor/pkg/util"
 	"github.com/sigstore/sigstore-go/pkg/root"
 	"github.com/sigstore/sigstore-go/pkg/tuf"
+	tlog "github.com/transparency-dev/formats/log"
 )
 
 // Default values for monitoring job parameters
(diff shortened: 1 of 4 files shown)


cmd/rekor_monitor/main.go
```diff
@@ -226,16 +242,34 @@ func mainLoopV2(tufClient *tuf.Client, flags *cmd.MonitorFlags, config *notifica
 				}
 			}
 
-			// TODO: should return prev and cur
-			cur, err := rekor_v2.RunConsistencyCheck(context.Background(), rekorShards, latestShardOrigin, flags.LogInfoFile)
+			prev, cur, err := rekor_v2.RunConsistencyCheck(context.Background(), rekorShards, latestShardOrigin, flags.LogInfoFile)
```
ret2libc, line 245: @facutuesca I think this was wrong, as `rekor_v2.RunConsistencyCheck` was returning just the previous checkpoint and not the current one, however here we were considering it as the current one.

Correct? Please have a deeper look to ensure what I did make sense.

facutuesca (approved):

ret2libc:
@fghanmi you are right, handling nil was removed while doing some PR refactoring. Thanks for catching that, I'm fixing it in https://github.com/sigstore/rekor-monitor/pull/746 .

PR #741, opened 2025-08-28
Add config validation

### LINKED ISSUE #737: Regex passed as input should be validated before being used
**Description**

Right now the various regex (e.g. for the cert subject) are compiled for each entry being analyzed, however we can compile and validate them early on, before the monitoring loop is even started, so to report a failure if they are wrong.

At the moment, instead, if you pass a wrong regex every entry is considered as errored.

ret2libc:
#### Summary
Closes #737 . Instead of failing to compile regular expressions only when they are used while matching identities (making the IdentitySearch process fail), this patch validate the configuration immediately after parsing it, to ensure it can be used correctly later on.

#### Release Note
* Add configuration validation immediately after parsing


mihaimaruseac (approved):
Oh, this is great! thank you

PR #749, opened 2025-09-02
internal/cmd: Use interface instead of callbacks

ret2libc:
I think this is more golang-like and it also ensures we don't accidentally forget to implement some callbacks in the struct (thing that happened quite a bit while developing).


mihaimaruseac (dismissed):
This is cleaner, thank you!

mihaimaruseac (approved):

Issue #378, opened 2024-04-25
Alert only for certificates issued from a set of trusted roots


Hayden-IO:
**Description**

Rekor accepts certificates from any issuer, including self-signed certificates. A malicious entity or spammer could issue certificates using someone's identity and OIDC issuer to trigger an alert for a monitor. This can be mitigated by verifying the certificate chains up to a trusted root. For the public instance, the monitor can pull in the trusted PKI from Sigstore's TUF repo. 

This should be configurable, since this may be used for self-hosted instances.

haydentherapper:
One comment from the linked thread on rekor is that it is possible to have a freeze attack against local metadata up to the expiration of the timestamp. This would delay fetching the latest trust root and could result in an entry being ignored. We could add a configuration to always fetch the latest TUF metadata regardless of timestamp if this is a concern based on the users threat model.

Horiodino:
Hey @haydentherapper, this issue looks quite interesting! Is it okay if I take it on? Let me know if there are any references or guidelines on where to make changes, and if the issue is still open.

haydentherapper:
Of course, thanks!

For some background info: Sigstore ships its [root of trusts](https://github.com/sigstore/root-signing/tree/main/targets) to verify certificates and log entries via [TUF](https://theupdateframework.io/).

I'd suggest first tackling https://github.com/sigstore/rekor-monitor/issues/51, to integrate a TUF client and fetch the log public keys in https://github.com/sigstore/rekor-monitor/blob/f4bd8b5532be295337f6ddfef41290e9cdb05e44/pkg/ct/consistency.go#L45 and https://github.com/sigstore/rekor-monitor/blob/f4bd8b5532be295337f6ddfef41290e9cdb05e44/pkg/rekor/verifier.go#L35. You'll see an example of a TUF client implemented in https://github.com/sigstore/sigstore-go/tree/main/pkg/tuf. That will make this issue smaller in scope.

This issue is specifically about narrowing down which certificates are verified from the log. The log contains both certificates issued from Sigstore's CA and from private CAs. When monitoring entries for specified IDs, users are going to only want to search for certificates that come from CAs they trust. Otherwise, someone could issue a certificate for "hayden@website" from their own private PKI, upload it to the log, and my monitor would alert.

[...]

PR #764, opened 2025-09-12
Add trusted CAs for filtering matched identities

ret2libc:
#### Summary
Partially addresses https://github.com/sigstore/rekor-monitor/issues/378 for rekor-monitor.

- [x] Consider TUF data when `-trusted-ca` is not specified
- [ ] Add support for CT monitor. It seems CTLog has a non regular x509 certificate, but it is built differently and I had an hard time trying to verifying it.: https://github.com/sigstore/rekor-monitor/pull/770

### DIFF
--- a/pkg/rekor/v1/identity.go
+++ b/pkg/rekor/v1/identity.go
@@ -157,7 +157,7 @@ func MatchedIndices(logEntries []models.LogEntry, mvs identity.MonitoredValues, 
 
 			// Validate that the certificate chain up to a trusted CA
 			if err := identity.ValidateCertificateChain(certs, trustedCAs); err != nil {
-				fmt.Fprintf(os.Stderr, "error validating certificate chain for UUID %s at index %d: %v\n", uuid, *entry.LogIndex, err)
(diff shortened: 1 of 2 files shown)


mihaimaruseac (approved):

ret2libc:
I'm also working on adding support for the CT log filtering, but it can also be done in a separate PR. It requires some more understanding to see what to validate exactly, because the precertificates are a bit special. Me and @facutuesca are working on that anyway.

pkg/rekor/v1/identity.go
```diff
@@ -154,6 +155,12 @@ func MatchedIndices(logEntries []models.LogEntry, mvs identity.MonitoredValues)
 				continue
 			}
 
+			// Validate that the certificate chain up to a trusted CA
+			if err := identity.ValidateCertificateChain(certs, trustedCAs); err != nil {
+				fmt.Fprintf(os.Stderr, "error validating certificate chain for UUID %s at index %d: %v\n", uuid, *entry.LogIndex, err)
```
haydentherapper, line 160: nit: Should we soften the language here, so it's not an error but an FYI?

haydentherapper (approved):
SGTM to handle CT in another PR

haydentherapper (approved):

PR #770, opened 2025-09-16
Check Certificate chain in CTLogs

ret2libc:
#### Summary
Check certificates in CTLogs are actually signed by trusted CAs.
Use separate roots/intermediates.

#### Release Note
* Add support for Certificates verification in CTLogs
* Split --trusted-ca flag into --ca-roots and --ca-intermediates, similar to cosign


cmd/ct_monitor/main.go
```diff
@@ -113,14 +115,14 @@ func (l CTMonitorLogic) GetEndIndex(cur cmd.LogInfo) *int64 {
 }
 
 func (l CTMonitorLogic) IdentitySearch(ctx context.Context, config *notifications.IdentityMonitorConfiguration, monitoredValues identity.MonitoredValues) ([]identity.MonitoredIdentity, []identity.FailedLogEntry, error) {
-	return ct.IdentitySearch(ctx, l.fulcioClient, *config.StartIndex, *config.EndIndex, monitoredValues, config.OutputIdentitiesFile, config.IdentityMetadataFile)
+	return ct.IdentitySearch(ctx, l.fulcioClient, *config.StartIndex, *config.EndIndex, monitoredValues, config.OutputIdentitiesFile, config.IdentityMetadataFile, config.CARoots, config.CAIntermediates)
```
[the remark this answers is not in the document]
facutuesca: most of the arguments are inside the `config`, maybe we can pass that instead?

mihaimaruseac (dismissed):

haydentherapper (approved):

pkg/identity/identity.go
```diff
@@ -429,42 +430,119 @@ func getSubjectAlternateNames[Certificate *x509.Certificate | *google_x509.Certi
...
-		return err
+	var roots *x509.CertPool
+	var err error
+	if caRoots != "" {
+		roots, err = getCertPool[*x509.CertPool](caRoots)
+		if err != nil {
+			return err
+		}
+	} else {
+		roots = nil
```
facutuesca, line 477: is this needed?

pkg/identity/identity.go
```diff
@@ -429,42 +430,119 @@ func getSubjectAlternateNames[Certificate *x509.Certificate | *google_x509.Certi
...
+	}
+
+	var intermediates *x509.CertPool
+	if caIntermediates != "" {
+		intermediates, err = getCertPool[*x509.CertPool](caIntermediates)
+		if err != nil {
+			return err
+		}
+	} else {
+		intermediates = nil
```
facutuesca, line 487: same here

pkg/identity/identity.go
```diff
@@ -429,42 +430,119 @@ func getSubjectAlternateNames[Certificate *x509.Certificate | *google_x509.Certi
...
+	}
+
+	for _, cert := range certs {
+		opts := google_x509.VerifyOptions{
+			DisableTimeChecks:              true,
+			DisableCriticalExtensionChecks: true,
+			DisableEKUChecks:               true,
+			DisablePathLenChecks:           true,
+			DisableNameConstraintChecks:    true,
+			DisableNameChecks:              false,
```
facutuesca, line 538: could we add comments here explaining why we use these options?

pkg/notifications/notifications.go
```diff
@@ -102,7 +102,28 @@ type IdentityMonitorConfiguration struct {
...
+	}
+	// Check if the file is a valid PEM file
+	caBytes, err := os.ReadFile(pemFile)
+	if err != nil {
+		return fmt.Errorf("failed to read trusted CA file %s: %v", pemFile, err)
+	}
+	block, _ := pem.Decode(caBytes)
+	if block == nil {
+		return fmt.Errorf("invalid trusted CA file %s: not a valid PEM file", pemFile)
+	}
```
facutuesca, line 125: we could propagate the decoding error to our error message here

pkg/ct/monitor.go
```diff
@@ -88,10 +91,28 @@ func ScanEntryOIDExtensions(logEntry ct.LogEntry, monitoredOIDMatchers []extensi
...
+func MatchedIndices(logEntries []ct.LogEntry, mvs identity.MonitoredValues, caRoots string, caIntermediates string) ([]identity.LogEntry, []identity.FailedLogEntry, error) {
 	matchedEntries := []identity.LogEntry{}
 	failedEntries := []identity.FailedLogEntry{}
 	for _, entry := range logEntries {
+		if entry.X509Cert != nil {
+			cert, err := x509.ParseCertificate(entry.X509Cert.Raw)
+			if err == nil {
+				if err = identity.ValidateCertificateChain([]*x509.Certificate{cert}, caRoots, caIntermediates); err != nil {
+					fmt.Fprintf(os.Stderr, "error validating certificate chain for log entry at index %d: %v\n", entry.Index, err)
+					continue
```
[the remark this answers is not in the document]
ret2libc: My understanding is that `failedEntries` are more for things that can't be parsed for whatever reason and ideally there should be none, however entries that are signed by another CA are possible.

If our purpose is to avoid spamming the user with notifications, then we should not add these entries to that list.

pkg/notifications/notifications.go
ret2libc: `func Decode(data []byte) (p *Block, rest []byte) {` the second return value is not error here

internal/cmd/common.go
```diff
@@ -161,53 +165,104 @@ func ParseAndLoadConfig(defaultServerURL, defaultTUFRepository, defaultOutputFil
...
+		fulcioCA := ca.(*root.FulcioCertificateAuthority)
+
+		if fulcioRootFile != nil {
+			// Get the root certificate from TUF
+			if err := pem.Encode(fulcioRootFile, &pem.Block{Type: "CERTIFICATE", Bytes: fulcioCA.Root.Raw}); err != nil {
+				if fulcioIntermediateFile != nil {
+					fulcioIntermediateFile.Close()
+					os.Remove(fulcioIntermediateFile.Name())
+				}
+				fulcioRootFile.Close()
```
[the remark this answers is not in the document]
facutuesca: (but yeah, this makes the code shorter but the behavior less explicit, so it's fine to leave as-is)

internal/cmd/common.go
```diff
@@ -197,72 +204,44 @@ func GetTUFClient(flags *MonitorFlags) (*tuf.Client, error) {
...
-						}
-						fulcioRootFile.Close()
-						os.Remove(fulcioRootFile.Name())
-						return nil, fmt.Errorf("failed to write Fulcio CA intermediate to temp file: %w", err)
-					}
-				}
-			}
+		// Get the root certificate from TUF
+		if err := pem.Encode(fulcioRootFile, &pem.Block{Type: "CERTIFICATE", Bytes: fulcioCA.Root.Raw}); err != nil {
+			cleanup()
```
facutuesca, line 235: won't this try to remove the files without closing them?

facutuesca (approved):
🚀

PR #771, opened 2025-09-17
Fix bug with missing entries when regex were used

### LINKED ISSUE #687: Identity monitoring fails to notify when monitored identities contain regex
**Description**

The way identity monitoring for CT and Rekor logs currently works is:
1. Gather all log entries that match any of the monitored identities (using [`rekor.IdentitySearch`](https://github.com/sigstore/rekor-monitor/blob/dbbfb0f5466c0f69f40995b81d7a2f6fe7b28950/pkg/rekor/identity.go#L285) for Rekor and  [`ct.MatchedIndices`](https://github.com/sigstore/rekor-monitor/blob/dbbfb0f5466c0f69f40995b81d7a2f6fe7b28950/pkg/ct/monitor.go#L71) for Fulcio)
2.  Create a list of monitored identities/values, where each monitored identity is a string (using [`identity.CreateIdentitiesList`](https://github.com/sigstore/rekor-monitor/blob/dbbfb0f5466c0f69f40995b81d7a2f6fe7b28950/pkg/identity/identity.go#L106))
    - Each string is the value of a configuration field (e.g.: an issuer, a cert subject, an extension value)
3. Create a map from monitored identity string (2) to list of log entries (1) using [`identity.CreateMonitoredIdentities`](https://github.com/sigstore/rekor-monitor/blob/dbbfb0f5466c0f69f40995b81d7a2f6fe7b28950/pkg/identity/identity.go#L127)
    - This is done by going through each log entry from (1), and checking if any of its values matches one of the string values from the list computed in (2).
4. (For Fulcio monitoring) Send notifications with the matching log entries, grouped by monitored value

There are a couple of problems with this:
- In (3), we check if any of the values in the entry matches one of the monitored values by using both as [keys in a map](https://github.com/sigstore/rekor-monitor/blob/dbbfb0f5466c0f69f40995b81d7a2f6fe7b28950/pkg/identity/identity.go#L130). This means that if the monitored value does not exactly match the entry value, we will not add the log entry to the map. 
    - For example, if the monitored value is `certSubject: \.*@mycompany\.com`, the monitored identity string in (2) will be `\.*@mycompany\.com`, but when used as a key in a map, it won't be the same as an actual value in the certificate (e.g: `name@mycompany\.com`)

- More generally, there's no domain separation between monitored values: (2) puts all values in a single list, no matter if they're certSubjects, issuers or a custom extension value. This means that if we monitor `issuer: google.com` and `extensionValues: google.com` (for some extension) as separate identities, entries that match either will be grouped under the same key `google.com`.

I think a solution would be to group the log entries by monitored identities in (1), since we're already doing the matching there. The grouping should not be by monitored value (e.g `google.com` in the example above), but by monitored identity (`issuer: google.com`). Also, by grouping them in (1), we don't need to match them a second time (like in (3)) in order to group them.

cc @haydentherapper

ret2libc:
#### Summary
Closes #687 .
Properly group matched log entries even when regex are used and matched.

#### Release Note
* Fix an implementation bug where identities matched through regex were not properly notified


facutuesca (approved):
🚀

Issue #829, opened 2025-12-03
ct_monitor: Hardcoded CT log key and nil log_id in STH result


fghanmi:
When testing `ct_monitor` against a freshly initialized custom CT log (and also against the public good instance `https://ctfe.sigstore.dev/2022`), there are two issues observed:
- In [getCTLogVerifier()](https://github.com/sigstore/rekor-monitor/blob/main/pkg/ct/consistency.go#L35), the `ctfe2022KeyID` is hardcoded. This causes the monitor to always use the Sigstore public good instance log key instead of the actual CT log being monitored (a custom CT log).
- The generated STH does not populate the `log_id` field correctly. The `log_id` field is 32 bytes of zero (`AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA=` in base64) (ie: `null`). I believe this can cause issues since the log_id should be SHA256(DER-encoded public key).

Example STH:
`{"sth_version":0,"tree_size":1,"timestamp":1764709491415,"sha256_root_hash":"DrRmmJyXZ/ohIyh5wHjAUYww+nO2yVXoXBHFP6voMIw=","tree_head_signature":"BAMARzBFAiEA+l8wzaH1ZS50rUJv6RcddzRsz1LXP4RddmGGFGK7ErcCIBTPnvQEvzw+fneBebiuGM0BtPs09GDfuuXlYzg71t67","log_id":"AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA="}`

**Expected Behavior:**
- `ct_monitor` should use the public key of the CT log being monitored, not a hardcoded `ctfe2022KeyID`.
- The `log_id` in the STH should correctly reflect the SHA-256 of the DER-encoded CT log public key, **if this is really intended**.

**Proposed Solution:**
Introduce a new optional CLI flag for `ct_monitor`:
`--ct_public_key=<PATH_TO_PEM>`

CC @ret2libc

ret2libc:
Related to the public key, what about getting the public key based on the URL in the trusted root? We already have the ct log URL.

fghanmi:
@ret2libc Yes, it can also solve both issues. Would you like that I prepare a PR for that ?

ret2libc:
cc @haydentherapper what do you think about the proposed approach?

fghanmi:
I've prepared a PR to give more details about the discussed solution: https://github.com/sigstore/rekor-monitor/pull/830

PR #830, opened 2025-12-05
Add dynamic ctlog verifier selection based on URL and correct STH log_id generation

fghanmi:
#### Summary

This update removes the previously hardcoded CTlog key ID and instead dynamically selects the correct CTlog verifier based on the CTlog's BaseURL. It also fixes the issue where STHs generated by the monitor contained an empty log_id field by correctly computing the log ID as the SHA-256 hash of the DER-encoded public key. These changes allow ct_monitor to work properly with both custom CT logs and the public Sigstore CT log instances.

This PR also includes a variable rename, `fulcioClient` -> `ctlogClient`

Issue: https://github.com/sigstore/rekor-monitor/issues/829

### DIFF
--- a/pkg/ct/consistency.go
+++ b/pkg/ct/consistency.go
@@ -113,6 +113,16 @@ func RunConsistencyCheck(logClient *ctclient.LogClient, logInfoFile string) (*ct
 		return nil, nil, fmt.Errorf("error fetching latest STH: %v", err)
 	}
 
+	// Ensure the verifier is always set if nil
+	if logClient.Verifier == nil {
+		verifier, err := getCTLogVerifier(logClient.BaseURI())
+
+		if err != nil {
+			return nil, nil, fmt.Errorf("error loading public key: %v", err)
+		}
+		logClient.Verifier = verifier
+	}
+
 	fi, err := os.Stat(logInfoFile)
 	// File containing previous checkpoints exists
 	var prevSTH *ct.SignedTreeHead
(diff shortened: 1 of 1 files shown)


pkg/ct/consistency.go
```diff
@@ -56,14 +67,14 @@ func getCTLogVerifier() (*ct.SignatureVerifier, error) {
 	return nil, fmt.Errorf("could not find certificate transparency log in trusted root")
 }
 
-func verifyCertificateTransparencyConsistency(logInfoFile string, logClient *ctclient.LogClient, signedTreeHead *ct.SignedTreeHead) (*ct.SignedTreeHead, error) {
+func verifyCertificateTransparencyConsistency(logInfoFile, serverURL string, logClient *ctclient.LogClient, signedTreeHead *ct.SignedTreeHead) (*ct.SignedTreeHead, error) {
```
cmurphy: `logClient` already has a `BaseURI()`, could that be used instead of a new parameter?

pkg/ct/consistency.go
```diff
@@ -109,7 +121,32 @@ func RunConsistencyCheck(logClient *ctclient.LogClient, logInfoFile string) (*ct
 		if err != nil {
 			return nil, nil, fmt.Errorf("error verifying consistency between previous and current STHs: %v", err)
 		}
+	} else if os.IsNotExist(err) {
+		if logClient.Verifier == nil {
```
haydentherapper: Can we move this out of this conditional and always set the verifier if it's unset? I think that looks a bit cleaner.

pkg/ct/consistency.go
```diff
@@ -109,7 +121,32 @@ func RunConsistencyCheck(logClient *ctclient.LogClient, logInfoFile string) (*ct
...
+			}
+			logClient.Verifier = verifier
+		}
+	}
+	if logClient.Verifier.PubKey != nil {
+		logID, err := ctLogIDFromPublicKey(logClient.Verifier.PubKey)
+		if err != nil {
+			return nil, nil, fmt.Errorf("error computing log ID: %v", err)
+		}
+		currentSTH.LogID = logID
```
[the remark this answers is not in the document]
haydentherapper: Ah I see, it's because of the fixed size array. Thanks. I don't think this has to be fixed as it doesn't impact verification, but I'm fine with fixing it to avoid it looking weird.

Issue #832, opened 2025-12-12
Questions/feedback for "GitHub workflow setup"


ericcornelissen:
I came here from [Catching malicious package releases using a transparency log](https://blog.trailofbits.com/2025/12/12/catching-malicious-package-releases-using-a-transparency-log/) and wanted to give the [GitHub Actions setup](https://github.com/sigstore/rekor-monitor/blob/bb4acc455862b723e58ecef1b34f92975362b37f/README.md#github-workflow-setup) a try, however I'm struggling a bit to understand everything. I'm by no means an expert in Rekor/Transparency logs, but I imagine you'd like this project to be used by people that aren't, so I don't think reporting these "shortcomings" should be considered unreasonable.

- What does the [Consistency check](https://github.com/sigstore/rekor-monitor/blob/bb4acc455862b723e58ecef1b34f92975362b37f/README.md#consistency-check) do? I _think_ it monitors the transparency log for entries corresponding the the repository in which the workflow is running, it would be nice to have that spelled out in the docs.
- What does it mean for there to be a [monitoring failure](https://github.com/sigstore/rekor-monitor/blob/bb4acc455862b723e58ecef1b34f92975362b37f/.github/workflows/reusable_monitoring.yml#L21)? Conceptually I understand but practically I don't know how the workflow would know when a transparency log entry is problematic...
- I'm afraid there's no satisfactory answer here but why does it need [`id-token: write`](https://github.com/sigstore/rekor-monitor/blob/bb4acc455862b723e58ecef1b34f92975362b37f/README.md?plain=1#L120) permission? The reason I ask is because it just feels very dangerous, consider: "Hey, just use this tool and we will monitor the transparency log for you. We just need your `id-token` that we can abuse and then neglect to inform you about it if we do."
- Please, please don't make me use [`@main`](https://github.com/sigstore/rekor-monitor/blob/bb4acc455862b723e58ecef1b34f92975362b37f/README.md?plain=1#L121). While I don't mind using the commit SHA, having actual human-readable and machine-usable version tags to refer to is, imho, a must.
- Not a big deal but does `file_issue` open an issue every time it detects a problem, or does it use one "status" issue that it updates and manages over time?
[...]

PR #1021, opened 2026-08-10
Replace `detect-workflow-js` action by `job` context

### LINKED ISSUE #1020: `slsa-framework/slsa-github-generator` is unmaintained
**Description**

This project uses [`slsa-framework/slsa-github-generator`](https://github.com/slsa-framework/slsa-github-generator) in various was, including in ways that affect users ([example](https://github.com/sigstore/rekor-monitor/blob/7dd89b58b660bef2bb05ba6ad977a09c85bb5140/.github/workflows/base_monitoring.yml#L84)). Unfortunately, it was [recently announced](https://github.com/slsa-framework/slsa-github-generator/commit/bb91a05077afa6601a3d7538c4cbbdbf0abe7ed9) that this project is "no longer actively maintained."

The main reason for opening this issue - besides the fact that migrating is simply the right thing to do - is that its usage is causing a warning with a deprecation warning:

> Node.js 20 is deprecated. The following actions target Node.js 20 but are being forced to run on Node.js 24: slsa-framework/slsa-github-generator/.github/actions/detect-workflow-js@f7dd8c54c2067bafc12ca7a55595d5ee9b75204a. For more information see: https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/

Meaning that, in the not-too-distant-future, this project will stop working as well (at least on public or up-to-date private runners).

ericcornelissen:
#### Summary

Remove the use of `slsa-framework/slsa-github-generator/.github/ actions/detect-workflow-js`, and by extension drop the need for the `id-token: write` permission, by the [`job` context](https://docs.github.com/en/actions/reference/workflows-and-actions/contexts#job-context). This context contains the information previously obtained with that action, without requiring a sensitive permission.

- This addresses the primary concern voiced in #832.
- This resolves #1020 by removing the dependency on `slsa-framework/slsa-github-generator` altogether.

#### Release Note

- Remove the `id-token: write` permission requirement from reusable workflows.

#### Documentation

I updated the docs in this project, I'm unaware of any docs on <https://docs.sigstore.dev> for this.

### DIFF
--- a/.github/workflows/base_monitoring.yml
+++ b/.github/workflows/base_monitoring.yml
@@ -58,18 +58,12 @@ on:
 
 permissions:
   contents: read
-  id-token: write
   issues: write
 
 jobs:
   detect-workflow:
     runs-on: ubuntu-latest
     timeout-minutes: 60
-    permissions:
-      id-token: write # Needed to detect the current reusable repository and ref.
-    outputs:
-      repository: ${{ steps.detect.outputs.repository }}
-      ref: ${{ steps.detect.outputs.ref }}
     steps:
       - name: Check monitor type
         run: |
@@ -79,9 +73,6 @@ jobs:
           fi
         env:
           INPUT_MONITOR_TYPE: ${{ inputs.monitor_type }}
-      - name: Detect the repository and ref
-        id: detect
-        uses: slsa-framework/slsa-github-generator/.github/actions/detect-workflow-js@f7dd8c54c2067bafc12ca7a55595d5ee9b75204a # v2.1.0
     # NOTE: This GHA should not be run concurrently.
     concurrency:
       group: ${{ inputs.monitor_type }}-consistency-check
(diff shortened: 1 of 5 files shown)


ericcornelissen:
As an aside I noticed that the old implementation has a TOCTOU-style bug where [the](https://github.com/sigstore/rekor-monitor/blob/2cbcfa260c42a9b2f060c7769adb5d1c31a3b7c7/.github/workflows/base_monitoring.yml#L98) [different](https://github.com/sigstore/rekor-monitor/blob/2cbcfa260c42a9b2f060c7769adb5d1c31a3b7c7/.github/workflows/base_monitoring.yml#L152) [jobs](https://github.com/sigstore/rekor-monitor/blob/2cbcfa260c42a9b2f060c7769adb5d1c31a3b7c7/.github/workflows/base_monitoring.yml#L171) may check out a different commit due to the use of a floating ref ([as documented](https://github.com/sigstore/rekor-monitor/blob/2cbcfa260c42a9b2f060c7769adb5d1c31a3b7c7/README.md?plain=1#L130)). Apart from some shenanigans or hickups, I don't think this bug can be used to achieve anything interesting which you couldn't already as a result of the use of a floating ref.

This fixes that because `${{ job.workflow_sha }}` is always a commit SHA, even when the the workflow is used as `sigstore/rekor-monitor/.github/workflows/reusable_monitoring.yml@main`.

ericcornelissen:
Btw, I have tested these changes over at https://github.com/ericcornelissen/playground-gha/actions/workflows/example.yml

Hayden-IO (approved):
Thanks so much!

PR #1023, opened 2026-08-10
Harden CI based on zizmor

ericcornelissen:
#### Summary

Harden the CI workflows, internal and external, in two key ways based on a [zizmor](https://docs.zizmor.sh/) scan:

- Consistently use `persist-credentials: false` per the [artipacked](https://docs.zizmor.sh/audits/#artipacked) rule. A hardening measure to avoid leaking credentials, already [in use once](https://github.com/sigstore/rekor-monitor/blob/672d3d16806a4b545e98d0b6535d0c2ea8a287ea/.github/workflows/release.yml#L45).
- Harden against [template injection](https://docs.zizmor.sh/audits/#template-injection). I opted to replace all uses of [GitHub Actions Expressions](https://docs.github.com/en/actions/reference/workflows-and-actions/expressions) (those `${{ }}` things) present in shell scripts by environment variables. This avoids any ambiguity about whether or not an attacker could control it and to what extend, just avoiding the issue altogether. This hardening is already in place for [one expression in a shell script](https://github.com/sigstore/rekor-monitor/blob/672d3d16806a4b545e98d0b6535d0c2ea8a287ea/.github/workflows/base_monitoring.yml#L68-L75).
  - In a few cases I went a bit beyond this mitigation and also wrapped the environment variable in double quotes to prevent [argument splitting](https://www.shellcheck.net/wiki/SC2086).

#### Release Note

NONE

#### Documentation

n/a

### DIFF
--- a/.github/workflows/base_monitoring.yml
+++ b/.github/workflows/base_monitoring.yml
@@ -87,6 +87,7 @@ jobs:
         with:
           repository: ${{ job.workflow_repository }}
           ref: "${{ job.workflow_sha }}"
+          persist-credentials: false
       - uses: actions/setup-go@b7ad1dad31e06c5925ef5d2fc7ad053ef454303e # v7.0.0
         with:
           go-version-file: './go.mod'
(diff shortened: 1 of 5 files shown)


.github/workflows/base_monitoring.yml
```diff
@@ -140,6 +156,7 @@ jobs:
         with:
           repository: ${{ job.workflow_repository }}
           ref: "${{ job.workflow_sha }}"
+          persist-credentials: false
```
Hayden-IO, line 159: Have you tested this out? I may be wrong, but I think we do need the credential to be stored in order to use it in the reporting script.

ericcornelissen: You're probably right.

I did test this successfully, but it didn't create (e.g.?) an issue so the credentials probably weren't needed. I'll test these changes with the reporting scripts properly asap.

ericcornelissen: Turns out it's not necessary to persist credentials here either, the reason being that the script uses the `GITHUB_TOKEN` ([example](https://github.com/sigstore/rekor-monitor/blob/672d3d16806a4b545e98d0b6535d0c2ea8a287ea/.github/workflows/scripts/report_failure.sh#L49)) whereas this option is about persisting the [git credentials used for git operations such as push](https://github.com/actions/checkout/blob/f548e57e544e1ff5a4c46bf1e1b8685f8e4a348a/README.md?plain=1#L112).

Tested in [this (simulated) failing](https://github.com/ericcornelissen/playground-gha/actions/runs/31534269092/job/93921735239) and [this succeeding](https://github.com/ericcornelissen/playground-gha/actions/runs/31534421784/job/93922160126) run operating on [this issue](https://github.com/ericcornelissen/playground-gha/issues/48).

mihaimaruseac (approved):

.github/workflows/base_monitoring.yml
Hayden-IO: Perfect, thank you for verifying this!

Hayden-IO (approved):

